PR #2, opened 2016-12-13
running localhost simulation

### LINKED ISSUE #1: Simulation-library
Now: onet-simulation is a standalone app
wanted: onet-simulation as a library

ineiti:
closes #1


simul/platform/deterlab.go
```diff
@@ -91,12 +98,16 @@ var simulConfig *onet.SimulationConfig
...
 	// Directory setup - would also be possible in /tmp
 	pwd, _ := os.Getwd()
-	d.cothorityDir = pwd + "/cothority"
-	d.deterDir = pwd + "/platform/deterlab"
-	d.deployDir = d.deterDir + "/remote"
-	d.buildDir = d.deterDir + "/build"
+	d.simulDir = pwd
+	d.deployDir = pwd + "/deploy"
+	d.buildDir = pwd + "/build"
+	_, file, _, _ := runtime.Caller(0)
```
nikkolasg: just to get it right, that means deleterlab will set the plartform path to the `main.go` file provided by the user in their own repo, is that right ?

simul/platform/deterlab.go
```diff
@@ -108,71 +119,60 @@ func (d *Deterlab) Configure(pc *Config) {
...
 
-	// start building the necessary packages
-	packages := []string{"simul", "users"}
-	if build != "" {
-		packages = strings.Split(build, ",")
+	// start building the necessary binaries - it's always the same,
+	// but built for another architecture.
+	packages := []pkg{
+		{"simul", "amd64", "linux", d.simulDir},
+		{"users", "386", "freebsd", path.Join(d.platformDir, "deterlab_users", ".")},
```
nikkolasg: so are we building the binary for both platforms now ? I thought only users was needed on the freebsd platform.

simul/simul.go
```diff
@@ -1,9 +1,26 @@
...
+
+Usually you start small, then work your way up to the full potential of your
+protocol!
 */
 package simul
+
+import "flag"
+
+// Start decides on the '--simul'-flag whether it needs to build or start the
+// simulation.
```
nikkolasg: I know I'am annoying but can you write down a quick sentence about if / when / how this flag should be given. Like `the flag --simul is usually given by the platform's implementation` or stg.

simul/platform/deterlab.go
ineiti: yes, so that it can re-compile that for the simulation-platform.

simul/platform/deterlab.go
ineiti: for deterlab we need two binaries:
- users - for users.deterlab.net
- simul (former cothority) - for the servers

simul/simul.go
ineiti: You need to be annoying, else I do stupid stuff ;) s/annoying/patient/.

simul/platform/deterlab.go
nikkolasg: +1

PR #9, opened 2016-12-14
Sda to onet

### LINKED ISSUE #8: Name-changes and network-output
some more name changes sda->onet
network-library error output removal in test
non-parallel test (until network-library and onet don't both use port 2000)

ineiti:
closes #8


Makefile
```diff
@@ -32,7 +32,7 @@ test_lint:
 test_playground:
 	cd services/skipchain; \
 	for a in $$( seq 10 ); do \
-	  go test -v -race -short || exit 1 ; \
+	  go test -v -race -short -p=1 || exit 1 ; \
```
nikkolasg: ? why go back to one processor ?

ineiti: because `onet` and `network` both use port `2000` :( It's related to https://github.com/dedis/cothority/issues/723 I think.

PR #18, opened 2016-12-15
[Network] Fixing panic memory invalid address

nikkolasg:
In encoding.go when it receives an unknown type, it calls `type.Name()` while the `type` is nil, because we wanted nice error output :(


ineiti:
+1 for adding a test ;)

PR #23, opened 2016-12-15
webhost server and client

### LINKED ISSUE #22: getWebHost has to return correct host for clients
now: getWebHost returns always host 0.0.0.0
wanted: getWebHost can return address for server and client

ineiti:
closes #22

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -55,7 +55,7 @@ func NewWebSocket(si *network.ServerIdentity) *WebSocket {
 		services:  make(map[string]Service),
 		startstop: make(chan bool),
 	}
-	webHost, err := getWebHost(si)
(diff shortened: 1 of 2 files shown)


websocket.go
```diff
@@ -357,10 +357,14 @@ func (ce *cerror) Error() string {
 }
 
 // getWebHost returns the host:port+1 of the serverIdentity.
-func getWebHost(si *network.ServerIdentity) (string, error) {
+func getWebHost(si *network.ServerIdentity, server bool) (string, error) {
```
nikkolasg: + `getWebAddress` seems more appropriate now, 
+ `server` => `global` 
+ A small comment saying what is this boolean doing.

PR #36, opened 2016-12-19
[Onet] One CreateProtocol for all

### LINKED ISSUE #19: Merge `CreateProtocolSDA` and `CreateProtocolService`
- `CreateProtocolSDA` - can we merge that even with `CreateProtocolService`?

nikkolasg:
This PR fixes #19 . Namely, it removes the `CreateProtocolOnet` and thus forces every `ProtocolInstance` creation to pass by the `CreateProtocol` method. Protocols or tests needs to pass the `NilServiceID`, and `Service` who does not want to impl. their `NewProtocol` method using `ServiceProcessor` has nothing to do.


overlay.go
```diff
@@ -568,8 +546,8 @@ func (o *Overlay) CreateProtocolService(name string, t *Tree, sid ServiceID) (Pr
 }
 
 // StartProtocol will create and start a P.I.
-func (o *Overlay) StartProtocol(t *Tree, name string) (ProtocolInstance, error) {
-	pi, err := o.CreateProtocolOnet(name, t)
+func (o *Overlay) StartProtocol(name string, t *Tree, sid ServiceID) (ProtocolInstance, error) {
+	pi, err := o.CreateProtocol(name, t, sid)
```
ineiti: +1 for unification in arguments.
Please add some more comment and complete "P.I." - a casual reader might not now what "P.I." is.

service.go
```diff
@@ -297,3 +301,22 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
 	}
 	return serv, true
 }
+
+// NewProtocol contains the logic of how and where a ProtocolInstance is
```
ineiti: lower-case "n".

service.go
```diff
@@ -297,3 +301,22 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
...
+// created. If the token's ServiceID is nil, then onet handles the creation of
```
ineiti: I propose to write "ONet" everywhere, what do you think?
Again, please don't use abbreviations in the comment for PI and other elements.

service.go
```diff
@@ -297,3 +301,22 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
...
+// the PI. If the corresponding service returns (nil,nil), then onet handles
+// the creation of the PI. Otherwise the service is responsible for setting up
+// the PI.
+func (s *serviceManager) newProtocol(tni *TreeNodeInstance, config *GenericConfig) (ProtocolInstance, error) {
+	si, ok := s.serviceByID(tni.Token().ServiceID)
+	defaultHandle := func() (ProtocolInstance, error) { return s.conode.protocolInstantiate(tni.Token().ProtoID, tni) }
+	if !ok {
+		// let onet handles it
```
ineiti: English grammar (EG): singular

service.go
```diff
@@ -297,3 +301,22 @@ func (s *serviceManager) serviceByID(id ServiceID) (Service, bool) {
...
+		return defaultHandle()
+	}
+	pi, err := si.NewProtocol(tni, config)
+	if pi == nil && err == nil {
+		pi, err = defaultHandle()
```
ineiti: why not `return defaultHandle()`?

service.go
nikkolasg: In the readme or presentations, I'm ok with "ONet", but in the code, I would stick with "onet" because 
+  it is easier to read & remember: which letter is uppercase? is it "Onet" or "ONet" or "ONET" ? 
+  to use this library, one must import "onet" and not "ONet"; it might feel strange to have two name for one thing. 

While I agree that ONet sounds more sexy and I am totally ok to show this library with such a name if we feel it's better, in the code, a more consistent and easier naming is probably better.

ineiti: OK, ONet for docu and onet for code.

ineiti:
special comment: we can take coveralls coverage as inceptive to add a test somewhere...

PR #58, opened 2017-01-09
[Network] #2 RegisterMessage 

nikkolasg:
This issue fixes https://github.com/dedis/cothority/issues/767 .
It removes some unused methods but mostly remove the duplicate and leaves on `RegisterMessage` as the only way to register a message to the network.


network/encoding.go
```diff
@@ -180,17 +169,12 @@ var globalOrder = binary.BigEndian
 // https://github.com/golang/go/commit/feb2a5d6103dad76b6374c5f346e33d55612cb2a
 var EmptyEnvelope = Envelope{MsgType: MessageTypeID(uuid.Nil)}
 
-// global mutex for MarshalRegisteredType
-var marshalLock sync.Mutex
```
ineiti: So you say there is no race-error if you remove that?

network/encoding.go
```diff
@@ -252,8 +236,8 @@ func UnmarshalRegistered(buf []byte) (MessageTypeID, Message, error) {
 	return tID, ptrVal.Interface(), nil
 }
 
-// MarshalBinary the application packet => to bytes
-// Implements BinaryMarshaler interface so it will be used when sending with protobuf
+// MarshalBinary returns the bytes representation of the envelope's message and
```
ineiti: wonderful genitive in `envelope's`

network/encoding.go
nikkolasg: At least, travis and me did not find anything.

network/encoding.go
nikkolasg: Thanks to your constant repetition, I think my head has spared some memory for that ... xD

PR #59, opened 2017-01-09
load save file and memory

ineiti:
Implements Cothority#491

### DIFF
--- a/context.go
+++ b/context.go
@@ -28,6 +28,10 @@ type Context struct {
 	network.Dispatcher
 }
 
+// ENVConodeData is the environmental variable that can be used to override the
+// service-data-path.
+const ENVConodeData = "CONODE_SERVICE_PATH"
+
 // defaultContext is the implementation of the Context interface. It is
 // instantiated for each Service.
 func newContext(c *Conode, o *Overlay, servID ServiceID, manager *serviceManager) *Context {
@@ -128,6 +132,8 @@ var contextData = map[string][]byte{}
 
 // Save takes an identifier and an interface. The interface will be network.Marshaled
 // and written to the filename based on the identifier. An eventual error will be returned.
+// If the destination is a file, it will be created with rw-r----- permissions
+// (0640).
 func (c *Context) Save(id string, data interface{}) error {
 	buf, err := network.MarshalRegisteredType(data)
 	if err != nil {
@@ -138,7 +144,7 @@ func (c *Context) Save(id string, data interface{}) error {
 		contextData[fname] = buf
 		return nil
 	}
-	return ioutil.WriteFile(fname, buf, 0660)
+	return ioutil.WriteFile(fname, buf, 0640)
 }
 
 // Load takes a filename and returns the network.Unmarshaled data. If an error
@@ -151,7 +157,7 @@ func (c *Context) Load(id string) (interface{}, error) {
 		var ok bool
 		buf, ok = contextData[c.Path(id)]
 		if !ok {
-			return nil, errors.New("This exntry doesn't exist")
+			return nil, errors.New("This entry doesn't exist")
 		}
 	} else {
 		var err error
@@ -174,26 +180,31 @@ func (c *Context) Path(id string) string {
 }
 
 // Returns the path to the file for storage/retrieval of the service-state.
-// It tries to come up with a suitable name for MacOS/Linux/Windows and also
-// accepts an override with the environmental-variable "CONODE_SERVICE".
+// It choses a suitable name for MacOS/Linux/Windows and also
+// accepts an override with the environmental-variable "CONODE_SERVICE_DATA".
+// Mac: ~/Library/Conode/Services
+// Other Unix: ~/.local/share/conode
+// Windows: $HOME$\AppData\Local\Conode
+// The directory will be created using rwxr-x--- permissions (0750).
 func initContextDataPath() {
 	// Set contextDataMemory to true if we're running in a test
-	p := os.Getenv("CONODE_SERVICE")
+	p := os.Getenv(ENVConodeData)
 	if p == "" {
 		u, err := user.Current()
 		if err != nil {
-			switch runtime.GOOS {
-			case "darwin":
-				p = path.Join(u.HomeDir, "Library", "Conode", "Services")
-			case "freebsd", "linux", "netbsd", "openbsd", "plan9", "solaris":
-				p = path.Join(u.HomeDir, ".local", "share", "conode")
-			case "windows":
-				p = path.Join(u.HomeDir, "AppData", "Local", "Conode")
-			default:
-				log.Fatal("Couldn't find OS")
-			}
+			log.Fatal("Couldn't get current user's environment:", err)
+		}
+		switch runtime.GOOS {
+		case "darwin":
+			p = path.Join(u.HomeDir, "Library", "Conode", "Services")
+		case "freebsd", "linux", "netbsd", "openbsd", "plan9", "solaris":
+			p = path.Join(u.HomeDir, ".local", "share", "conode")
+		case "windows":
+			p = path.Join(u.HomeDir, "AppData", "Local", "Conode")
+		default:
+			log.Fatal("Couldn't find OS")
 		}
 	}
-	os.MkdirAll(p, 0770)
+	os.MkdirAll(p, 0750)
 	contextDataPath = p
 }
--- a/context.go
+++ b/context.go
@@ -133,15 +133,16 @@ func (c *Context) String() string {
 var contextData = map[string][]byte{}
 
 // Save takes an identifier and an interface. The interface will be network.Marshaled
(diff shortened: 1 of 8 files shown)


context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	_, ret, err := network.UnmarshalRegistered(buf)
+	return ret, err
+}
+
+// Path returns the absolute path to load and save the configuration.
+// The file is chosen as "#{ServerIdentity.Public}_#{ServiceName}_#{id}.bin",
```
nikkolasg: The id is generated out of the service name no ? I mean I think these two informations are redundant.

context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+// so no service and no conode share the same file.
+func (c *Context) Path(id string) string {
+	pub, _ := c.ServerIdentity().Public.MarshalBinary()
+	return path.Join(contextDataPath, fmt.Sprintf("%x_%s_%s.bin", pub,
+		ServiceFactory.Name(c.ServiceID()), id))
+}
+
+// Returns the path to the file for storage/retrieval of the service-state.
+// It tries to come up with a suitable name for MacOS/Linux/Windows and also
+// accepts an override with the environmental-variable "CONODE_SERVICE".
```
nikkolasg: `CONODE_SERVICE_PATH` ?
Or something that indicates the purpose. 
And also, it should be put into a global variable, since you refer to it multiple times here and in the test.

context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+// The file is chosen as "#{ServerIdentity.Public}_#{ServiceName}_#{id}.bin",
+// so no service and no conode share the same file.
+func (c *Context) Path(id string) string {
+	pub, _ := c.ServerIdentity().Public.MarshalBinary()
+	return path.Join(contextDataPath, fmt.Sprintf("%x_%s_%s.bin", pub,
+		ServiceFactory.Name(c.ServiceID()), id))
+}
+
+// Returns the path to the file for storage/retrieval of the service-state.
+// It tries to come up with a suitable name for MacOS/Linux/Windows and also
```
nikkolasg: The comment should describe what these suitables names are if they exists OR put it in the README somewhere.

context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+// accepts an override with the environmental-variable "CONODE_SERVICE".
+func initContextDataPath() {
+	// Set contextDataMemory to true if we're running in a test
+	p := os.Getenv("CONODE_SERVICE")
+	if p == "" {
+		u, err := user.Current()
```
nikkolasg: ẁhat happens if there is an error ? 
Should we panic ? Should we default to current directory ? Or ... ?

context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+		if err != nil {
```
nikkolasg: Should it not be `if err == nil ` ?? I.e. if there is not error retrieving the user base directory, I compute the full path... ?

And stylistic nitpick after that: Should prefer reduce the indentation by putting the "quit-the-flow condition" inside the if
```
if err != nil {
  // what todo !?
  panic(err)
}
switch runtime.GOOS {
...
}
```

context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+			case "freebsd", "linux", "netbsd", "openbsd", "plan9", "solaris":
+				p = path.Join(u.HomeDir, ".local", "share", "conode")
+			case "windows":
+				p = path.Join(u.HomeDir, "AppData", "Local", "Conode")
+			default:
+				log.Fatal("Couldn't find OS")
+			}
+		}
+	}
+	os.MkdirAll(p, 0770)
```
nikkolasg: Some unclear moving parts:
+ Why not tighter restrictions by default ? like 0600 or 0644 ?
+ if the directory already exists the permissions won't be changed neither checked
+ it's missing a comment saying that the `CONODE_SERVICE` folder will be created if it does not exists already

context.go
ineiti: Except if a service wants to save more than one piece of data. But we could say that it only saves one chunk of data, even thought that's a little restriction...

context.go
```diff
@@ -96,3 +123,77 @@ func (c *Context) Service(name string) Service {
...
+	buf, err := network.MarshalRegisteredType(data)
+	if err != nil {
+		return err
+	}
+	fname := c.Path(id)
+	if contextDataPath == "" {
+		contextData[fname] = buf
+		return nil
+	}
+	return ioutil.WriteFile(fname, buf, 0660)
```
nikkolasg: Permissions should be specified within the comments, and I'd put it as 0640 or stg. I don't know it's a delicate subject...

context.go
```diff
@@ -23,6 +40,16 @@ func newContext(c *Conode, o *Overlay, servID ServiceID, manager *serviceManager
 	}
 }
 
+// contextDataPath indicates where the service-data will be stored. If it is
+// empty, a memory-map is used.
+var contextDataPath = ""
```
nikkolasg: Just for the record, I'm ok now with this custom rule of being empty now, but we'll have to change that once we have a config file / struct in place. Like a real interface `Memory`, `File`, `UploadAmazon` whatever. :D

context.go
ineiti: Oups - bad me...

context.go
ineiti: a directory needs the execute-bit set, else you cannot `cd` inside. So it's either 0700 or 0770.

Higher up you see the file is 0660. Personally I never use groups in a unix-system, but some people do, and then it might be useful to allow for other people to be able to access that configuration, too.

context.go
ineiti: I put 0750 for directory and 0640 for the file.

context.go
ineiti: When the time comes, we can add that.

context.go
```diff
@@ -96,3 +127,84 @@ func (c *Context) Service(name string) Service {
...
+		ServiceFactory.Name(c.ServiceID()), id))
+}
+
+// Returns the path to the file for storage/retrieval of the service-state.
+// It choses a suitable name for MacOS/Linux/Windows and also
+// accepts an override with the environmental-variable "CONODE_SERVICE_DATA".
+// Mac: ~/Library/Conode/Services
+// Other Unix: ~/.local/share/conode
+// Windows: $HOME$\AppData\Local\Conode
+// The directory will be created using rwxr-x--- permissions (0750).
```
nikkolasg: `... if it does not exists. Otherwise, the directory is left unchanged, including the permissions.`

context.go
```diff
@@ -23,6 +44,16 @@ func newContext(c *Conode, o *Overlay, servID ServiceID, manager *serviceManager
 	}
 }
 
+// contextDataPath indicates where the service-data will be stored. If it is
+// empty, a memory-map is used.
+var contextDataPath = ""
+
+func init() {
+	if flag.Lookup("test.v") == nil {
```
nikkolasg: Mhhhhh... so what happens when a test is not ran with the verbose flag ?

I'm not a fan of this custom if here but yet again, I don't know of any other easy solutions.... :/

ineiti: It doesn't matter, because even without the verbose flag, "test.v" is defined, but it is `false`. First time I like _go_ making the difference between `nil` and `false` ;)

context.go
```diff
@@ -96,3 +129,99 @@ func (c *Context) Service(name string) Service {
 func (c *Context) String() string {
 	return c.conode.ServerIdentity.String()
 }
+
+var contextData = map[string][]byte{}
+
+// Save takes an identifier and an interface. The interface will be network.Marshaled
+// and written to the filename based on the identifier. An eventual error will be returned.
+// If the destination is a file, it will be created with rw-r----- permissions
+// (0640).
```
nikkolasg: `If the file already exists, it will be *truncated*`

context.go
```diff
@@ -96,3 +129,99 @@ func (c *Context) Service(name string) Service {
...
+		}
+	}
+	_, ret, err := network.UnmarshalRegistered(buf)
+	return ret, err
+}
+
+// Path returns the absolute path to load and save the configuration.
+// The file is chosen as "#{ServerIdentity.Public}_#{ServiceName}_#{id}.bin",
+// so no service and no conode share the same file.
+func (c *Context) Path(id string) string {
```
nikkolasg: I'm wondering: since we forbid/remove the ability to have a directory for a service now, is this necessary to put it as a public method ?  
If not, then the comments are still good, but maybe should be put at the top of the file so it's shown in the doc. 
If yes, then do we assume that a Service can use this path ? If so, it is a problem or not regarding the loading/saving methods ?

PR #64, opened 2017-01-11
[Network] #5 Proposition - Message ID uint16

nikkolasg:
This PR is a proposal to change our current way of registering messages to the network library. Before, when registering a message, the network library had a very specific, non customizable, and non portable way of computing a message ID (the id that uniquely identify a message), and the format was a long(ish) uuid of 16 bytes.  The whole problem has been described intensively here: https://github.com/dedis/cothority/issues/561#issuecomment-244051327

With this PR, the ID is now a regular uint32 (smallest unit supported by protobuf), and it's directly decided by the user upon registration of a message.  The wire protocol now looks like:
```
Size (uint32) || ID (uint32) || Message ([]bytes) 
```
Each module should pick a range of ids to register their packets. The registration is now looking like this:
```
var myCustomID network.MessageID = 34

func init() {
    network.RegisterMessage(myCustomID,myCustomMessage{})
}
```
It panics if it registers a different packets under the same id, so any error should be detected at init time, assuming the registration happens in the "init" func.

It's a proposal, and I'd like inputs on this one, mostly on how to properly manage the ids: how to make sure collisions are avoided... I've done it now before v1 because it changes the external API.


nikkolasg:
After discussion with @bford , let's postpone the refactoring of the wire protocol for later.

PR #69, opened 2017-01-12
only app-changes

ineiti:
Closes cothority-773

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -27,10 +27,10 @@ import (
 )
 
 // DefaultServerConfig is the default server configuration file-name.
-const DefaultServerConfig = "config.toml"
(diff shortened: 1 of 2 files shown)


app/server.go
```diff
@@ -27,10 +27,10 @@ import (
 )
 
 // DefaultServerConfig is the default server configuration file-name.
-const DefaultServerConfig = "config.toml"
+const DefaultServerConfig = "private.toml"
 
 // DefaultGroupFile is the default group definition file-name.
-const DefaultGroupFile = "group.toml"
+const DefaultGroupFile = "public.toml"
```
nikkolasg: the beauty of putting such thing in variables appears now !

PR #76, opened 2017-01-13
added proto file

### LINKED ISSUE #75: proto-file for network
Copy proto-file for network here.

ineiti:
Closes #75

### DIFF
--- a/network/network.proto
+++ b/network/network.proto
@@ -1,0 +1,12 @@
+message ServerIdentity{
+    required bytes public = 1;
+    required bytes id = 2;
+    required string address = 3;
+    optional string description = 4;
+}
+
+message Roster {
+    optional bytes id = 1;
+    repeated ServerIdentity list = 2;
+    required bytes aggregate = 3;
+}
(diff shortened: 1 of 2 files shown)


network/network.proto
```diff
@@ -0,0 +1,12 @@
+message ServerIdentity{
+    required bytes public = 1;
+    required bytes id = 2;
+    required string address = 3;
+    optional string description = 4;
+}
+
+message Roster {
+    optional bytes id = 1;
```
nikkolasg: 1. Roster should be in Onet
2. ID is optional or required ? To me, it should be optional as it's something that can be generated but I don't really insist on it. However, it should be consistent with ServerIdentity no ? If it's required there, it should be required here ?

network/network.proto
```diff
@@ -0,0 +1,10 @@
+package network
```
ineiti: is that `network` or `onet.network`?

nikkolasg: What do you prefer ? I prefer to keep it simple and put it independant of onet, just "network" but as it does not matter much in the wire protocol, I don't have a strong opinion on this.

PR #78, opened 2017-01-13
[Onet] Reducing unused exported API

nikkolasg:
This PR is a first shot to reducing the exported API that we have in Onet and which is not used at all elsewhere, and which should not be used at all elsewhere.
If a method has been privatized, it's because it's building everywhere, including cothority (i.e. cothority does not use it).

**Note**: One last thing I'd like to reduce is the `CloseHost` method on `TreeNodeInstance` which basically shutdowns everything... It's bad ! Unfortunatly it's used in the simulation,which is in a diferent package. Any ideas ?

### DIFF
--- a/local.go
+++ b/local.go
@@ -174,7 +174,7 @@ func (l *LocalTest) CloseAll() {
 	}
 	for _, node := range l.Nodes {
 		log.Lvl3("Closing node", node)
-		node.Close()
+		node.closeDispatch()
 	}
 	l.Nodes = make([]*TreeNodeInstance, 0)
 	// Give the nodes some time to correctly close down
--- a/local.go
+++ b/local.go
@@ -181,8 +181,8 @@ func (l *LocalTest) CloseAll() {
 	//time.Sleep(time.Millisecond * 500)
 }
 
-// GetTree returns the tree of the given TreeNode
-func (l *LocalTest) GetTree(tn *TreeNode) *Tree {
+// getTree returns the tree of the given TreeNode
+func (l *LocalTest) getTree(tn *TreeNode) *Tree {
 	var tree *Tree
 	for _, t := range l.Trees {
 		if tn.IsInTree(t) {
@@ -199,7 +199,7 @@ func (l *LocalTest) NewTreeNodeInstance(tn *TreeNode, protName string) (*TreeNod
 	if o == nil {
 		return nil, errors.New("Didn't find corresponding overlay")
 	}
-	tree := l.GetTree(tn)
+	tree := l.getTree(tn)
 	if tree == nil {
 		return nil, errors.New("Didn't find tree corresponding to TreeNode")
 	}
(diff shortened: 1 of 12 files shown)


local.go
```diff
@@ -174,15 +174,15 @@ func (l *LocalTest) CloseAll() {
...
 	}
 	l.Nodes = make([]*TreeNodeInstance, 0)
 	// Give the nodes some time to correctly close down
 	//time.Sleep(time.Millisecond * 500)
 }
 
-// GetTree returns the tree of the given TreeNode
-func (l *LocalTest) GetTree(tn *TreeNode) *Tree {
+// getTree returns the tree of the given TreeNode
+func (l *LocalTest) getTree(tn *TreeNode) *Tree {
```
ineiti: localtest is used for local tests, so it's OK to export these functions.

local.go
```diff
@@ -220,18 +220,18 @@ func (l *LocalTest) NewTreeNodeInstance(tn *TreeNode, protName string) (*TreeNod
 	return node, nil
 }
 
-// GetNodes returns all Nodes that belong to a treeNode
-func (l *LocalTest) GetNodes(tn *TreeNode) []*TreeNodeInstance {
+// getNodes returns all Nodes that belong to a treeNode
+func (l *LocalTest) getNodes(tn *TreeNode) []*TreeNodeInstance {
```
ineiti: localtest is used for local tests, so it's OK to export these functions.

local.go
```diff
@@ -220,18 +220,18 @@ func (l *LocalTest) NewTreeNodeInstance(tn *TreeNode, protName string) (*TreeNod
...
 		nodes = append(nodes, n)
 	}
 	return nodes
 }
 
-// SendTreeNode injects a message directly in the Overlay-layer, bypassing
+// sendTreeNode injects a message directly in the Overlay-layer, bypassing
 // Host and Network
-func (l *LocalTest) SendTreeNode(proto string, from, to *TreeNodeInstance, msg network.Message) error {
+func (l *LocalTest) sendTreeNode(proto string, from, to *TreeNodeInstance, msg network.Message) error {
```
ineiti: localtest is used for local tests, so it's OK to export these functions.

local.go
```diff
@@ -245,16 +245,16 @@ func (l *LocalTest) SendTreeNode(proto string, from, to *TreeNodeInstance, msg n
 	return to.overlay.TransmitMsg(onetMsg, io)
 }
 
-// AddPendingTreeMarshal takes a treeMarshal and adds it to the list of the
+// addPendingTreeMarshal takes a treeMarshal and adds it to the list of the
 // known trees, also triggering dispatching of onet-messages waiting for that
 // tree
-func (l *LocalTest) AddPendingTreeMarshal(c *Server, tm *TreeMarshal) {
+func (l *LocalTest) addPendingTreeMarshal(c *Server, tm *TreeMarshal) {
```
ineiti: localtest is used for local tests, so it's OK to export these functions.

local.go
```diff
@@ -245,16 +245,16 @@ func (l *LocalTest) SendTreeNode(proto string, from, to *TreeNodeInstance, msg n
...
 	c.overlay.addPendingTreeMarshal(tm)
 }
 
-// CheckPendingTreeMarshal looks whether there are any treeMarshals to be
+// checkPendingTreeMarshal looks whether there are any treeMarshals to be
 // called
-func (l *LocalTest) CheckPendingTreeMarshal(c *Server, el *Roster) {
+func (l *LocalTest) checkPendingTreeMarshal(c *Server, el *Roster) {
```
ineiti: localtest is used for local tests, so it's OK to export these functions.

nikkolasg: This method is really internal to Overlay and should not be exported. It was in the first place because we had external tests so we needed to export API.

local.go
nikkolasg: This method is really internal to Overlay and should not be exported. It was in the first place because we had external tests so we needed to export API.

PR #80, opened 2017-01-13
merging only relevant changes from debugging_cothority764

ineiti:
Closes https://github.com/dedis/cothority/issues/764
And another bug I saw in Travis...

### DIFF
--- a/treenode.go
+++ b/treenode.go
@@ -286,7 +286,11 @@ func (n *TreeNodeInstance) closeDispatch() error {
 		n.msgDispatchQueueWait <- true
 	}
 	n.msgDispatchQueueMutex.Unlock()
-	return n.ProtocolInstance().Shutdown()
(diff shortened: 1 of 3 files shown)


treenode.go
```diff
@@ -286,7 +286,11 @@ func (n *TreeNodeInstance) closeDispatch() error {
 		n.msgDispatchQueueWait <- true
 	}
 	n.msgDispatchQueueMutex.Unlock()
-	return n.ProtocolInstance().Shutdown()
+	pni := n.ProtocolInstance()
+	if pni == nil {
```
nikkolasg: That's something very disturbing as there should be no `TreeNodeInstance` living without having a `ProtocolInstance` attached to it.
Was there an error with it ? Was it only in the tests ?

ineiti: was in a test where the system started to set up and then has been teared down before it was properly set up.

PR #87, opened 2017-01-17
Same origin 86

### LINKED ISSUE #86: Allow same origin to fail
The websocket-port should allow same origin to be false. This is because the script contacting the websocket will not be served from the conode, so same origin cannot be checked.

ineiti:
Closes #86 
Needs fix from #88 to pass checks.


nikkolasg:
Travis Fails.

PR #90, opened 2017-01-17
[Websocket] Graceful shutdown working

### LINKED ISSUE #17: Check if `graceful` is needed in `websocket`
I thought I need `websocket`, so we can close the server correctly, but perhaps not?

nikkolasg:
This PR fixes #17 : it makes possible to have a clean shutdown of a conode's websocket server while being to stop it correctly for the tests.


nikkolasg:
I've tested by building cothority/conode and I could run it then stop the server by ^C

PR #91, opened 2017-01-23
Adds ErrorHandler accessors

lbarman:
Gives the service the opportunity to register a error-handler function (type `func(*ServerIdentity)`). This function will be called on network error: Timeout, EOF, or ConnectionClosed, with the identity of the faulty remote server.

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -44,7 +44,7 @@ type Router struct {
 	// Every handler in this list is called by this router when a network error occurs (Timeout, Connection
 	// Closed, or EOF). Those handler should be added by using SetErrorHandler(). The 1st argument is the remote
 	// server with whom the error happened
-	connectionsErrorHandler []func(*ServerIdentity)
+	connectionErrorHandlers []func(*ServerIdentity)
 
 	// keep bandwidth of closed connections
 	traffic counterSafe
@@ -58,7 +58,7 @@ func NewRouter(own *ServerIdentity, h Host) *Router {
 		connections:             make(map[ServerIdentityID][]Conn),
 		host:                    h,
 		Dispatcher:              NewBlockingDispatcher(),
-		connectionsErrorHandler: make([]func(*ServerIdentity), 0),
+		connectionErrorHandlers: make([]func(*ServerIdentity), 0),
 	}
 	r.address = h.Address()
 	return r
@@ -207,7 +207,7 @@ func (r *Router) removeConnection(si *ServerIdentity, c Conn) {
 
 // triggerConnectionErrorHandlers trigger all registered connectionsErrorHandlers
 func (r *Router) triggerConnectionErrorHandlers(remote *ServerIdentity) {
-	for _, v := range r.connectionsErrorHandler {
+	for _, v := range r.connectionErrorHandlers {
 		v(remote)
 	}
 }
@@ -373,5 +373,5 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
 // on network error (e.g. Timeout, Connection Closed, or EOF) with the identity of the faulty
 // remote host as 1st parameter.
 func (r *Router) AddErrorHandler(errorHandler func(*ServerIdentity)) {
-	r.connectionsErrorHandler = append(r.connectionsErrorHandler, errorHandler)
+	r.connectionErrorHandlers = append(r.connectionErrorHandlers, errorHandler)
 }
--- a/network/router.go
+++ b/network/router.go
@@ -361,5 +361,8 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
 // on network error (e.g. Timeout, Connection Closed, or EOF) with the identity of the faulty
 // remote host as 1st parameter.
 func (r *Router) SetErrorHandler(errorHandler func(*ServerIdentity)) {
+	if errorHandler != nil {
+		panic("error handler was already set, cannot replace")
+	}
 	r.errorHandler = errorHandler
 }
--- a/network/router.go
+++ b/network/router.go
@@ -40,6 +40,10 @@ type Router struct {
 
 	// wg waits for all handleConn routines to be done.
 	wg sync.WaitGroup
+
+	//if not nil, is called by this entity when any network error occurs (e.g. TCP stream broken, host disconnected, etc)
+	//should be set by using SetErrorHandler(). The 1st argument is the remote server with whom the error happened
+	errorHappened func(*ServerIdentity)
 }
 
 // NewRouter returns a new Router attached to a ServerIdentity and the host we want to
@@ -220,12 +224,18 @@ func (r *Router) handleConn(remote *ServerIdentity, c Conn) {
 		if err != nil {
 			if err == ErrTimeout {
 				log.Lvlf5("%s drops %s connection: timeout", r.ServerIdentity.Address, remote.Address)
+				if r.errorHappened != nil {
+					r.errorHappened(remote)
+				}
 				return
 			}
 
 			if err == ErrClosed || err == ErrEOF {
 				// Connection got closed.
 				log.Lvlf5("%s drops %s connection: closed", r.ServerIdentity.Address, remote.Address)
+				if r.errorHappened != nil {
+					r.errorHappened(remote)
+				}
 				return
 			}
 			// Temporary error, continue.
@@ -345,4 +355,11 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
 	}
 	log.Lvl4(r.address, "Identity received from", dst.Address)
 	return dst, nil
+}
+
+// SetErrorHandler sets the network error handler function for this entity. The function will be called
+// on network error (e.g. TCP broken stream, host disconnected, etc) with the identity of the faulty
+// remote host as 1st parameter.
+func (r *Router) SetErrorHandler(errorHandler func(*ServerIdentity)) {
+	r.errorHappened = errorHandler
 }
--- a/network/router.go
+++ b/network/router.go
@@ -41,9 +41,10 @@ type Router struct {
 	// wg waits for all handleConn routines to be done.
 	wg sync.WaitGroup
 
-	//if not nil, is called by this entity when any network error occurs (e.g. Timeout, Connection Closed, or EOF)
-	//should be set by using SetErrorHandler(). The 1st argument is the remote server with whom the error happened
-	errorHandler func(*ServerIdentity)
+	// Every handler in this list is called by this router when any network error occurs (Timeout, Connection
+	// Closed, or EOF). Those handler should be added by using SetErrorHandler(). The 1st argument is the remote
+	// server with whom the error happened
+	errorHandlers []func(*ServerIdentity)
 }
 
 // NewRouter returns a new Router attached to a ServerIdentity and the host we want to
@@ -54,6 +55,7 @@ func NewRouter(own *ServerIdentity, h Host) *Router {
 		connections:    make(map[ServerIdentityID][]Conn),
 		host:           h,
 		Dispatcher:     NewBlockingDispatcher(),
+		errorHandlers:  make([]func(*ServerIdentity), 0),
 	}
 	r.address = h.Address()
 	return r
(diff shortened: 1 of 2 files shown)


network/router.go
```diff
@@ -40,6 +40,10 @@ type Router struct {
 
 	// wg waits for all handleConn routines to be done.
 	wg sync.WaitGroup
+
+	//if not nil, is called by this entity when any network error occurs (e.g. TCP stream broken, host disconnected, etc)
```
ineiti: Spaces between `//` and text.

ineiti: What is `entity` here?

ineiti: Give a list of which errors are propagated.

network/router_test.go
```diff
@@ -61,6 +61,13 @@ func testRouter(t *testing.T, fac routerFactory) {
 	case <-time.After(500 * time.Millisecond):
 		t.Fatal("TcpHost should have returned from Run() by now")
 	}
+
+	require.Nil(t, h.errorHappened)
+	errHandler := func(remote *ServerIdentity) {
+		//nothing to do here
+	}
+	h.SetErrorHandler(errHandler)
+	require.NotNil(t, h.errorHappened)
```
ineiti: Is it difficult to get a simulation for an error here? Like closing the connection, and check whether the error is propagated?

network/router.go
```diff
@@ -40,6 +40,10 @@ type Router struct {
...
+	//should be set by using SetErrorHandler(). The 1st argument is the remote server with whom the error happened
+	errorHappened func(*ServerIdentity)
```
ineiti: rename to `errorHandler`?

ineiti:
Usually we do:
1. open an issue
2. create a branch that ends with `_#issue`
3. write a short description at the beginning of the PR, including `Closes #issue` or `fixes #issue`

For now, can you add a short description at the beginning of the PR stating what problem you're solving?

nikkolasg:
Sorry but I can't resist proposing an alternative solution here:
What about geneic event registration and dispatching ? Something in the spirit of the system used in [`mobile`](https://godoc.org/golang.org/x/mobile/app).
```go
// in event.go
--------------------------------------------------------------------------------

type Event interface{}

type EventHandler interface {
    OnEvent(e Event)
}
// list of eventHandlers
var eventsHandler []EventHandler

// register an eventhandler so new event are dispatched to it
func RegisterEventHandler(eh EventHandler) { ... }

// dispatch a new event to all event handlers
func DispatchEvent(e Event) { ... }

// router.go
--------------------------------------------------------------------------------

type DropConnection struct {
   From ServerIdentity
   UnsentMessage Packet // for example
}
 ...
if err == ErrTimeout || err = ErrClosed {
	event.DispatchEvent(&DropConnection{remote,msg})
}

// let's say prifi.go 
--------------------------------------------------------------------------------

func (s *prifiService) OnEvent(e Event) {
   switch ev := e.(type) {
       case network.DropConnection:
          log.Lvl1f("Dropped Connection from %s (msg %s not sent)",ev.From,ev.UnsentMessage.String())
       default:
             // don't handle any other events
    }
}
```

This is something that can get very useful for a lot of other stuff as well, such as `NewConn` when a new connection is made, `NewTreeEvent` when receiving a new tree, `NewCoSiDone` when a CoSi has been done, `NewCiscJoined` when a server is responsible for a new identity "data-structure". Maybe for logging and statistics etc.

Maybe this is a bit over the edge and I certainly don't expect any big welcome to this idea here and I think it's something that can be relatively easy to do and very powerful at the same time. That's it. 
I'm still gonna comment on the PR anyway.

network/router.go
nikkolasg: `connDropped` ? `errorHappened` is like `whatever thing is wrong I put it here`. It does not *mean* much. Feel free to propose other names.

network/router.go
```diff
@@ -346,3 +356,10 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
 	log.Lvl4(r.address, "Identity received from", dst.Address)
 	return dst, nil
 }
+
+// SetErrorHandler sets the network error handler function for this entity. The function will be called
+// on network error (e.g. TCP broken stream, host disconnected, etc) with the identity of the faulty
+// remote host as 1st parameter.
+func (r *Router) SetErrorHandler(errorHandler func(*ServerIdentity)) {
+	r.errorHappened = errorHandler
```
nikkolasg: that means only one "errorHandler" can be registered at the same time?
Not a fan of the idea.
It would be preferable if it could keep a *list* of errorHandler and dispatch to all of them when the connection drops out. It should not be too much to do I guess.

network/router_test.go
nikkolasg: Test also whether the thing really works or not. If an a connection drops out, do you receive something on the handler?

ineiti:
That looks like a really great feature for V2 ;) Two things that make me prefer Ludo's solution:
- it's smaller - less possibility of doing it wrong
- we don't need more than it does for the moment

network/router.go
ineiti: or at least test if another handler is already registered and panic if it is.

network/router_test.go
```diff
@@ -63,6 +63,52 @@ func testRouter(t *testing.T, fac routerFactory) {
...
+	go h1.Start()
+	go h2.Start()
+
+	defer func() {
+		h1.Stop()
+	}()
+
+	// tests the setting error handler
+	require.Nil(t, h1.errorHandler)
+	errHandlerCalled := false
```
ineiti: I'd make this a channel, else you could get a race. Then you can `select` on the channel and add a timeout: https://gobyexample.com/timeouts

network/router_test.go
lbarman: Done

network/router.go
lbarman: Well it's `SetErrorHandler` not `AddErrorHandler`. A setter should not panic if called twice in my opinion

network/router_test.go
ineiti: I'm proud to have seen this before travis did ;)

network/router.go
nikkolasg: Well, I would still prefer multiple registration possible otherwise it does not make sense, you would have to write in the comments "Only the last registered error handler is the one to receive the error" .. -_-" 
For the cost of adding a slice, I find that very desirable...

network/router_test.go
lbarman: Indeed, done

network/router_test.go
```diff
@@ -63,6 +63,55 @@ func testRouter(t *testing.T, fac routerFactory) {
...
+	assert.Equal(t, 3, decoded.I)
+	assert.Nil(t, h2.Send(h1.ServerIdentity, msgSimple))
+	decoded = <-proc.relay
+
+	//stop node 2
+	h2.Stop()
+
+	// test if the error handler was called
+	select {
+	case _ = <-errHandlerCalled:
```
ineiti: `case <-errHandlerCalled:`

network/router_test.go
lbarman: Haha, I forgot the `-race`. I'm happy with cothority's Makefile, where `make test` does linting and testing with race etc...

nikkolasg:
> That looks like a really great feature for V2 ;) 
Why not for v1 ? It's something that add a new feature, it does not break API or something and is needed now.

ineiti:
LGTM, what's still missing:
- description in PR
- either panic or use a slice in `SetErrorHandler`

lbarman:
It it was my code, I wouldn't panic on `SetErrorHandler`, for reasons mentionned in the comment above. But OK

ineiti:
It's just in case two people want to add a handler and wonder why one doesn't work. @nikkolasg suggestion with a slice is cleaner, but adding a `panic` at least shows that there is an error.

lbarman:
Ah true, it could be used with multiple services.

network/router_test.go
```diff
@@ -63,6 +63,61 @@ func testRouter(t *testing.T, fac routerFactory) {
...
+	err := h1.Send(h2.ServerIdentity, msgSimple)
+	require.Nil(t, err)
+	decoded := <-proc.relay
+	assert.Equal(t, 3, decoded.I)
+	assert.Nil(t, h2.Send(h1.ServerIdentity, msgSimple))
+	decoded = <-proc.relay
+
+	//stop node 2
+	h2.Stop()
+	time.Sleep(250 * time.Millisecond)
```
nikkolasg: why do you sleep here ?
If it's because it takes time to detect the error and propagate the error to the handler, you should be using channels like below:
```go
select {
    case <-errHandlerCalled:
       // all good
    case <-time.After(100 * time.Millisecond):
      t.Error("Error handler should ... ")
}
```

network/router.go
```diff
@@ -40,6 +40,11 @@ type Router struct {
 
 	// wg waits for all handleConn routines to be done.
 	wg sync.WaitGroup
+
+	// Every handler in this list is called by this router when any network error occurs (Timeout, Connection
+	// Closed, or EOF). Those handler should be added by using SetErrorHandler(). The 1st argument is the remote
+	// server with whom the error happened
+	errorHandlers []func(*ServerIdentity)
```
nikkolasg: Please, change the name to reflect exactly the underlying meaning (here and in every other references).
> `any network error occurs `

That's wrong, there could be a listener error, there could be a file limit exhaustion error etc etc. 
`ConnDropped` or something else explicit.

network/router.go
```diff
@@ -220,12 +226,18 @@ func (r *Router) handleConn(remote *ServerIdentity, c Conn) {
 		if err != nil {
 			if err == ErrTimeout {
 				log.Lvlf5("%s drops %s connection: timeout", r.ServerIdentity.Address, remote.Address)
+				for _, v := range r.errorHandlers {
```
nikkolasg: can you put that into a small `Router` function please, let's avoid duplication as much as possible.

network/router.go
```diff
@@ -346,3 +358,10 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
 	log.Lvl4(r.address, "Identity received from", dst.Address)
 	return dst, nil
 }
+
+// AddErrorHandler adds a network error handler function for this router. The functions will be called
+// on network error (e.g. Timeout, Connection Closed, or EOF) with the identity of the faulty
+// remote host as 1st parameter.
+func (r *Router) AddErrorHandler(errorHandler func(*ServerIdentity)) {
```
ineiti: To make it complete, we'd need a `RemoveErrorHandler`.

network/router.go
ineiti: Something like `connectionErrorHandlers`?

network/router.go
nikkolasg: Do we ? Is it a use case to remove it at some point ? We did not do it for `Services`, for `Protocols` we only added it months later, etc...  I'm asking because it brings the trouble of finding the right one to delete. Just pointer comparison ?

network/router.go
nikkolasg: Yep, even if it's a big longish, I totally prefer that.

network/router.go
ineiti: A use case that comes into my mind is tests. But in fact you generate a new `Router` for each test anyway, so no, it's not needed. But I'll add it to your issue.

network/router_test.go
lbarman: Thanks for the tip, done

network/router.go
```diff
@@ -199,6 +205,13 @@ func (r *Router) removeConnection(si *ServerIdentity, c Conn) {
 	r.connections[si.ID] = arr[:len(arr)-1]
 }
 
+// triggerConnectionErrorHandlers trigger all registered connectionsErrorHandlers
+func (r *Router) triggerConnectionErrorHandlers(remote *ServerIdentity) {
```
ineiti: I think `connectionErrorHandlers` is the correct way of writing it, while `connectionsErrorHandler` puts the plural at the wrong place. So:
can you rename `connectionsErrorHandler` to `connectionErrorHandlers`?

lbarman: Done

ineiti: Nice - you just don't know how such small things throw me off when I read code :)

PR #95, opened 2017-01-24
Sum up Rx/Tx correctly in `network.Router`

### LINKED ISSUE #93: Router doesn't sum up tx/rx correctly
Now: the `onet.Router` only sums up rx/tx of active connections.
Wanted: if a connection stops, add it's rx & tx to a struct-global variable.

ineiti:
Closes #93

### DIFF
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -7,5 +7,5 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	log.MainTest(m, 4)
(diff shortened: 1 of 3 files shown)


nikkolasg:
Sorry I should have told you earlier but we have this nice `counterSafe` available in `struct.go`, could be nice to use it instead of re-inventing the wheel. :)

PR #100, opened 2017-01-26
Mininet 85

### LINKED ISSUE #85: Mininet
Port mininet from cothority-repo to onet.

ineiti:
Closes #85

### DIFF
--- a/simul/platform/.gitignore
+++ b/simul/platform/.gitignore
@@ -1,2 +1,0 @@
-mininet/
-localhost/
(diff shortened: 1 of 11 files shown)


simul/platform/.gitignore
```diff
@@ -1,2 +0,0 @@
-mininet/
```
nikkolasg: ? Sure about putting some "binary" folders inside our git ? Or do I misunderstand ?

ineiti: this used to be a binary-folder, now it's the `start.py` and `setup*.sh`-folder.

ineiti:
yes, as long as the servers are in the same subnet.

PR #101, opened 2017-01-26
TLS communication

### LINKED ISSUE #83: Add TLS-host
Now that we have a nice router-layer, add a TLS host for full TLS connections.
- [x] - implement TLS-host
- [x] - adjust apps for using TLS-host
- [ ] - cleanup

ineiti:
Closes #83 - comments from previous, cothority-PR:

- [x] - move `TLSKC.Key` from `ServerIdentity` to `TLSHost`
- [x] - check why it doesn't work with IP-addresses (127.0.0.1 instead of 'localhost').
- [ ] - clean up `CommonName` / `IP` / `DNS`
- [ ] - enter Country, Org, ... on keyboard
- [ ] - also allow `PlainTCP` in a `TLSHost` - does that really make sense?

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -22,6 +22,8 @@ import (
 	// CoSi-protocol is not part of the cothority.
 
 	// For the moment, the server only serves CoSi requests
+	"math/big"
+
 	"github.com/dedis/crypto/abstract"
 	crypconf "github.com/dedis/crypto/config"
 )
@@ -76,7 +78,7 @@ func InteractiveConfig(binaryName string) {
 		portStr = port
 	}
 
-	serverBinding = network.NewTCPAddress(hostStr + ":" + portStr)
+	serverBinding = network.NewTLSAddress(hostStr + ":" + portStr)
 	if !serverBinding.Valid() {
 		log.Error("Unable to validate address given", serverBinding)
 		return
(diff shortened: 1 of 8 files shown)


ineiti:
@nikkolasg:
`app.InteractiveConfig`:
* have the possibility of setting their own value for the fields instead of "ch" and "epfl"
* or give a path to an already PEM encoded certificate + key

network/tlskc.go
```diff
@@ -0,0 +1,128 @@
...
+func NewTLSCert(serial *big.Int, country, org, orgUnit string,
+	validYear int, subjectKeyID []byte,
+	ips []net.IP) *x509.Certificate {
+	return &x509.Certificate{
+		SerialNumber: serial,
+		Subject: pkix.Name{
+			Country:            []string{country},
+			Organization:       []string{org},
+			OrganizationalUnit: []string{orgUnit},
+			CommonName:         "*",
```
ineiti: @liamsi:
Both should be possible for the user to set: DNSNames and IPAddresses

app/server.go
```diff
@@ -133,12 +135,19 @@ func InteractiveConfig(binaryName string) {
 
 	// create the keys
 	privStr, pubStr := createKeyPair()
+	ips, err := net.LookupIP(publicAddress.Host())
+	log.ErrFatal(err)
+	cert, key, err := network.NewCertKey(network.NewTLSCert(big.NewInt(1), "ch", "epfl", "dedis",
```
ineiti: @nikkolasg:
* have the possibility of setting their own value for the fields instead of "ch" and "epfl"
* or give a path to an already PEM encoded certificate + key

PR #105, opened 2017-01-31
Makes app.Group's Description field public

lbarman:
Makes app.Group's Description field public. There is no constructor, and I need to set the description file for the simulations.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -88,12 +88,12 @@ type ServerToml struct {
 // Group holds the Roster and the server-description.
 type Group struct {
 	Roster      *onet.Roster
-	description map[*network.ServerIdentity]string
+	Description map[*network.ServerIdentity]string
 }
 
 // GetDescription returns the description of a ServerIdentity.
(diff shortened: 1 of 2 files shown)


ineiti:
so it was 4 lines after all ;)

app/config.go
```diff
@@ -88,12 +88,12 @@ type ServerToml struct {
 // Group holds the Roster and the server-description.
 type Group struct {
 	Roster      *onet.Roster
-	description map[*network.ServerIdentity]string
+	Description map[*network.ServerIdentity]string
 }
 
 // GetDescription returns the description of a ServerIdentity.
```
nikkolasg: Damn. I could have said a useful comment: 
We should add `this is kept for compatibility reasons for the moment` since basically this method is now completely useless.

lbarman: Too late 0:)

lbarman: Is it really though ? I guess PriFi is not the only protocol where nodes are not all alike

PR #107, opened 2017-02-01
[Crypto] Moving to v0

### LINKED ISSUE #106: [Crypto] Move to v0
Onet should be using the v0 branch of dedis/crypto so it stays stable while the refactoring is happening in dedis/crypto.

nikkolasg:
This PR fixes #106 : it moves all dependencies from `github.com/dedis/crypto` to `gopkg.in/dedis/crypto.v0`.

### DIFF
--- a/simulation.go
+++ b/simulation.go
@@ -10,10 +10,10 @@ import (
 	"net"
 
 	"github.com/BurntSushi/toml"
-	"github.com/dedis/crypto/abstract"
-	"github.com/dedis/crypto/config"
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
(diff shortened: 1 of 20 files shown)


simulation.go
```diff
@@ -264,7 +264,7 @@ func (s *SimulationBFTree) CreateRoster(sc *SimulationConfig, addresses []string
 		if sc.PrivateKeys[entities[0].Address].Equal(
```
ineiti: non-PR related question: do we want to keep this?

nikkolasg: I don't really understand this line actually... Was it just the "check" to make sure we were not using the same key for all nodes ? 
Why do you think it is not needed anymore ?

PR #115, opened 2017-02-07
adding all missing functions

ineiti:
Now we have all of
`(Read|Write)(64|Hex)(Scalar|Pub)`
and
`String(64|Hex)To(Scalar|Pub)`
and
`(Scalar|Pub)ToString(64|Hex)`

For V2: replace `Pub` with `Point`, see #116

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -5,10 +5,12 @@ import (
 	"encoding/hex"
 	"io"
 
+	"errors"
+
 	"gopkg.in/dedis/crypto.v0/abstract"
 )
 
-// Read64Pub a public point to a base64 representation
+// Read64Pub reads a public point from a base64 representation
 func Read64Pub(suite abstract.Suite, r io.Reader) (abstract.Point, error) {
 	public := suite.Point()
 	dec := base64.NewDecoder(base64.StdEncoding, r)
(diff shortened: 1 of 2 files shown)


crypto/key.go
```diff
@@ -55,6 +102,24 @@ func StringHexToPub(suite abstract.Suite, s string) (abstract.Point, error) {
...
+// String64ToPub reads a base64 representation of a public point and converts it
+// back to a point.
+func String64ToPub(suite abstract.Suite, s string) (abstract.Point, error) {
+	encoded, err := base64.StdEncoding.DecodeString(s)
+	if err != nil {
+		return nil, err
+	}
+	point := suite.Point()
+	err = point.UnmarshalBinary(encoded)
+	return point, err
```
nikkolasg: I think it can be reduced to one line with https://golang.org/pkg/strings/#NewReader
```
return Read64Pub(suite,strings.NewReader(s))
```
:)

crypto/key.go
```diff
@@ -72,9 +137,50 @@ func StringHexToScalar(suite abstract.Suite, str string) (abstract.Scalar, error
...
+
+// String64ToScalar reads a scalar in base64 from a string
+func String64ToScalar(suite abstract.Suite, str string) (abstract.Scalar, error) {
+	enc, err := base64.StdEncoding.DecodeString(str)
+	if err != nil {
+		return nil, err
+	}
+	s := suite.Scalar()
+	err = s.UnmarshalBinary(enc)
+	return s, err
```
nikkolasg: Also one line with 
```
return Read64Scalar(suite,strings.NewReader(str))
```

crypto/key.go
```diff
@@ -72,9 +137,50 @@ func StringHexToScalar(suite abstract.Suite, str string) (abstract.Scalar, error
...
+		return nil, errors.New("didn't get enough bytes from stream")
+	}
+	_, err = hex.Decode(bufByte, bufHex)
+	if err != nil {
+		return nil, err
+	}
+	return bufByte, nil
+}
+
+func writeHex(suite abstract.Suite, wc io.WriteCloser, data ...interface{}) error {
```
nikkolasg: is it used somewhere ? I can't find references in this PR at least.

crypto/key.go
ineiti: why is there no `NewWriter`?

crypto/key.go
ineiti: that was just an exercise to see if you read the code - <blush>

crypto/key.go
ineiti: If you have a simplification for the `getHex`, I'm all for it...

PR #117, opened 2017-02-08
websocket: add Tx / Rx method to websocket client

### LINKED ISSUE #96: [WebSocket] Client not implementing `monitor.CounterIOMeasure`
The `Client` from the websocket file does not implement the `monitor.CounterIOMeasure`. As a consequence, we can not measure the bandwidth during simulations from `Client`s using websocket.

nikkolasg:
It fixes #96.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -169,6 +169,8 @@ type Client struct {
 	conn    *websocket.Conn
 	// whether to keep the connection
 	keep bool
+	rx   uint64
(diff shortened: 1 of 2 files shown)


ineiti:
And why did you _not_ use `counterSafe` here?

nikkolasg:
because counterSafe is private in network and it was not needed since there is already a general lock in the client's structure.

PR #133, opened 2017-02-21
added changelog and a reference in the README

### LINKED ISSUE #132: add a CHANGELOG
We need to have a CHANGELOG with the most important changes that might break things...

ineiti:
Closes #132


nikkolasg:
We really need to do something about that changelog so it's automatically updated or at least in a convenient way for us...

PR #138, opened 2017-02-28
Simulation now reads the servers mapping from a config file, instead of having hardcoded values.

lbarman:
Previously, the simulation on deterlab would use hardcoded values for server names, of the form
`server-XXX.EXPERIMENT.PROJECT.isi.deterlab.net`
`10.255.0.X`

Depending on the topology, this is really inconvenient. For two LANs for instance, we need different subsets of IPs, and server names are useful to differentiates machines.

### DIFF
--- a/simul/platform/deterlab.go
+++ b/simul/platform/deterlab.go
@@ -371,10 +371,12 @@ func (d *Deterlab) createHosts() {
 	} else {
 		log.Lvl3("Found", ServersListTomlFile, ", using it for assigning IPs to servers")
 
+		numServers := d.Servers
+
 		// the simulation's main toml file requests a number of server (i.e. Servers=X). we need our mapping
 		// to have at least that many servers obviously; more isn't a problem, they won't be used.
-		if len(servers.Servers) < d.Servers {
-			log.Fatal("Simulation's toml specifies", d.Servers, "servers, but", ServersListTomlFile,
+		if len(servers.Servers) < numServers {
+			log.Fatal("Simulation's toml specifies", numServers, "servers, but", ServersListTomlFile,
 				"specifies only", len(servers.Servers), " servers.")
 		}
 
--- a/simul/platform/deterlab.go
+++ b/simul/platform/deterlab.go
@@ -425,7 +425,7 @@ func decodeServersTomlFile(filePath string) (*ServersListToml, error) {
 	defer f.Close()
 
 	servers := &ServersListToml{}
-	_, err = toml.DecodeReader(f, hosts)
+	_, err = toml.DecodeReader(f, servers)
 	if err != nil {
 		return nil, err
 	}
--- a/simul/platform/deterlab.go
+++ b/simul/platform/deterlab.go
@@ -34,6 +34,7 @@ import (
 
 	"path/filepath"
 
+	"errors"
 	"github.com/BurntSushi/toml"
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/log"
@@ -98,8 +99,8 @@ const ServersListTomlFile = "servers.toml"
 // Server is a mapping "deterlab machine name" -> physical IP address
 // example : server-0.EXP.PROJECT.isi.deterlab.net -> 10.1.2.3
 type Server struct {
-	Name     string
-	IP      string
+	Name string
+	IP   string
 }
 
 // ServersListToml is a list of mappings "deterlab machine name" -> physical
@@ -360,6 +361,40 @@ func (d *Deterlab) Wait() error {
 // Write the hosts.txt file automatically
 // from project name and number of servers
 func (d *Deterlab) createHosts() {
+
+	//try to open the file; if it fail, fallback to hard-coded mode
+	servers, err := decodeServersTomlFile(ServersListTomlFile)
+
+	if err != nil {
+		log.Error("Could not find or open", ServersListTomlFile, ", falling back to hard-coded values")
+		d.createDefaultHosts()
+	} else {
+		log.Lvl3("Found", ServersListTomlFile, ", using it for assigning IPs to servers")
+
+		// the simulation's main toml file requests a number of server (i.e. Servers=X). we need our mapping
+		// to have at least that many servers obviously; more isn't a problem, they won't be used.
+		if len(servers.Servers) < d.Servers {
+			log.Fatal("Simulation's toml specifies", d.Servers, "servers, but", ServersListTomlFile,
+				"specifies only", len(servers.Servers), " servers.")
+		}
+
+		d.Phys = make([]string, 0, numServers)
+		d.Virt = make([]string, 0, numServers)
+
+		for _, host := range servers.Servers {
+			d.Phys = append(d.Phys, host.Name)
+			d.Virt = append(d.Virt, host.IP)
+		}
+
+		log.Lvl3("Physical:", d.Phys)
+		log.Lvl3("Internal:", d.Virt)
+
+	}
+}
+
+// creates hosts name of the form server-X.EXP.PROJECT.isi.deterlab.net,
+// and IPs of the form 10.255.0.X
+func (d *Deterlab) createDefaultHosts() {
 	numServers := d.Servers
 
 	ip := "10.255.0."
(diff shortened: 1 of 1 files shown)


lbarman:
API was not changed, behavior not changed if `servers.toml` not present (except one harmless `log.Error` that warns about the fallback into hardcoded mode)
It would be great to have this into v1 when you have the time ;) thanks

ineiti:
Why use a `.toml`-file here? I think a file in the `/etc/hosts`-format would be much more convenient to write and parse:

```
IP1 hostname1
IP2 hostname2
```

Even better would be to use deterlab's commands to get those values directly - probably a combination of `node_list` and `expinfo` should do the trick.

ineiti:
```
[ineiti@users ~]$ expinfo -l -e SAFER,LB-LLD
Experiment: SAFER/LB-LLD
State: active

Virtual Lan/Link Info:
ID              Member/Proto    IP/Mask         Delay     BW (Kbs)  Loss Rate
--------------- --------------- --------------- --------- --------- ---------
lanclients      client-0:0      10.0.1.1        5.00      100000    0.00000000
                ethernet        255.255.255.0   5.00      100000    0.00000000
lanclients      client-1:0      10.0.1.2        5.00      100000    0.00000000
                ethernet        255.255.255.0   5.00      100000    0.00000000
lanclients      client-2:0      10.0.1.3        5.00      100000    0.00000000
                ethernet        255.255.255.0   5.00      100000    0.00000000
lanclients      client-3:0      10.0.1.4        5.00      100000    0.00000000
                ethernet        255.255.255.0   5.00      100000    0.00000000
lanclients      client-4:0      10.0.1.5        5.00      100000    0.00000000
                ethernet        255.255.255.0   5.00      100000    0.00000000
lanclients      relay:0         10.0.1.254      5.00      100000    0.00000000
                ethernet        255.255.255.0   5.00      100000    0.00000000
lantrustees     relay:1         10.1.0.254      50.00     100000    0.00000000
                ethernet        255.255.255.0   50.00     100000    0.00000000
lantrustees     trustee-0:0     10.1.0.1        50.00     10000     0.00000000
                ethernet        255.255.255.0   50.00     10000     0.00000000
lantrustees     trustee-1:0     10.1.0.2        50.00     10000     0.00000000
                ethernet        255.255.255.0   50.00     10000     0.00000000
lantrustees     trustee-2:0     10.1.0.3        50.00     10000     0.00000000
                ethernet        255.255.255.0   50.00     10000     0.00000000

Physical Lan/Link Mapping:
ID              Member          IP              MAC                  NodeID   
--------------- --------------- --------------- -------------------- ---------
lanclients      client-0:0      10.0.1.1        a0:36:9f:08:54:da    cpc7     
                                                4/1 <-> 6/17         HP2e4    
lanclients      client-1:0      10.0.1.2        a0:36:9f:09:27:fa    cpc86    
                                                4/1 <-> 7/24         HP2e3    
lanclients      client-2:0      10.0.1.3        a0:36:9f:08:58:26    cpc33    
                                                4/1 <-> 8/23         HP2e2    
lanclients      client-3:0      10.0.1.4        00:0e:0c:65:e0:31    bpc040   
                                                0/1 <-> 4/8          Bhpod    
lanclients      client-4:0      10.0.1.5        00:15:17:5d:32:a4    bpc174   
                                                0/1 <-> 6/7          Bhp3     
lanclients      relay:0         10.0.1.254      00:04:23:bb:25:aa    bpc083   
                                                0/1 <-> 8/11         Bhpmd    
lantrustees     relay:1         10.1.0.254      00:04:23:bb:25:ab    bpc083   
                                                1/1 <-> 4/4          Bhpmd    
lantrustees     trustee-0:0     10.1.0.1        a0:36:9f:09:27:dc    cpc38    
                                                2/1 <-> 6/20         HP2e2    
lantrustees     trustee-1:0     10.1.0.2        a0:36:9f:09:28:f6    cpc13    
                                                4/1 <-> 4/11         HP2e4    
lantrustees     trustee-2:0     10.1.0.3        d8:9d:67:ee:ff:71    hpc032   
                                                5/1 <-> 1/23         HPS9e1   

Virtual Queue Info:
ID              Member          Q Limit    Type    weight/min_th/max_th/linterm
--------------- --------------- ---------- ------- ----------------------------
lanclients      client-0:0      100 slots  Tail    0/0/0/0
lanclients      client-1:0      100 slots  Tail    0/0/0/0
lanclients      client-2:0      100 slots  Tail    0/0/0/0
lanclients      client-3:0      100 slots  Tail    0/0/0/0
lanclients      client-4:0      100 slots  Tail    0/0/0/0
lanclients      relay:0         100 slots  Tail    0/0/0/0
lantrustees     relay:1         100 slots  Tail    0/0/0/0
lantrustees     trustee-0:0     100 slots  Tail    0/0/0/0
lantrustees     trustee-1:0     100 slots  Tail    0/0/0/0
lantrustees     trustee-2:0     100 slots  Tail    0/0/0/0
```

ineiti:
In general I don't like to use a config-file for things that can be done automatically, which is the case here.

And yes, there are two IP-addresses for the relay, so we need to think how to do that correctly.

It's true that it doesn't change existing code, but if people start to rely on that code and we want to change it later, it will be difficult. This is why I'm hesitating to accept that solution.

And toml against `/etc/hosts` - to me the `toml`-way of defining an array just looks bad ;)

@nikkolasg - any comments?

lbarman:
OK I see. Do you find any case where my code is not flexible enough, e.g. we could not implement the topology with servers.toml ?
I had a second field IP2 in the toml, removed since it had no impact in the code, but my way would also support multiple IP's (at least for defining the servers)

nikkolasg:
Good things you found the command `expinfo -l -e SAFER,LB-LLD`, we tried searching with @lbarman but did not find how to get the ip addresses. 
It seems this commands finds everything deterlab needs to know to build the correct roster, doesn't it ?
@lbarman Why do you need two IPs for the relay ? I understand it's because the relay is part of two different networks, but what happens if the relay has only one ip address (from one network or the other) ? 
For example: 
```
network 1: 10.0.0.0/24  relay: 10.0.0.1
network 2: 10.0.1.0/24
```
If a node from network2 contacts the relay at `10.0.0.1`, what happens ? I figure the relay should be the main gateway of the network2 so the packet should flow towards the relay **as being the default route**, and the relay knows its own IP address.
If that turns out to be true **and** that the delay imposed in the network is respected, then I'm more inclined to do that automatically with the scripts deterlab provides. 
As in Matrix: `never send a human to do a machine's job`.

If it turns out to not work out, then I would be OK with the current suggestion. Some note:
 - how can you support multiple ip per node ? multiple entries in the list with the same name ? what happens in that case ? two entries in the roster ? This seems like a weird corner case we should make sure to understand before accepting anything.
 - add a flag for the default file location
 - I'm more opting to keep toml file also, instead of writing our own parser again, even if very simple... If you wish to go with JSON because arrays are nicer, why not, but TOML looks very fine to me for a configuration file.

lbarman:
So if I change for this: 
```
        command := fmt.Sprintf("/usr/testbed/bin/expinfo -l -e %s,%s", d.Project, d.Experiment)
	ans, err := SSHRun(d.Login, d.Host, command)
	if err != nil {
		log.Fatal("Error while discovering servers:", err)
	}
        // extract the interesting bit
	a := strings.Split(str, "Link Info:\n")
	c := strings.Split(a[1], "\n\n")
	d := strings.Split(c[0], "---------\n")
	interestingBit := d[1]
	lines := strings.Split(interestingBit, "\n")

	// extract one line over two
	interestingLines := make([]string, 0)
	for i, line := range lines {
		if i%2 == 0 {
			interestingLines = append(interestingLines, line)
		}
	}

	regex := regexp.MustCompile("([^ ]+)([ ]+)([^ ]+)([ ]+)([^ ]+)([ ]+)([^ ]+)([ ]+)([^ ]+)")

	// split by space and trim
	for i := range interestingLines {

		matches := regex.FindStringSubmatch(interestingLines[i])
		//lan := matches[1]
		nameInterface := matches[3]
		ip := matches[5]

		fmt.Println(nameInterface, "->", ip)
	}
```
would it be better ?

lbarman:
See https://github.com/dedis/onet/pull/139

PR #139, opened 2017-02-28
Simulation now reads the servers mapping from Deterlab's API, instead of having hardcoded values.

lbarman:
Previously, the simulation on deterlab would use hardcoded values for server names, of the form
`server-XXX.EXPERIMENT.PROJECT.isi.deterlab.net`
`10.255.0.X`

Depending on the topology, this is really inconvenient. For two LANs for instance, we need different subsets of IPs, and server names are useful to differentiates machines.

### DIFF
--- a/simul/platform/deterlab.go
+++ b/simul/platform/deterlab.go
@@ -28,7 +28,6 @@ import (
 	"strconv"
 	"time"
 
-	"errors"
 	"os/user"
 
 	"runtime"
--- a/simul/platform/deterlab.go
+++ b/simul/platform/deterlab.go
@@ -372,24 +372,18 @@ func (d *Deterlab) createHosts() {
 		}
 	}
 
-	// format : LAN	name:interface	ip	[otherdata]
-	// number of spaces is varying
-	regex := regexp.MustCompile("^([^ ]+)([ ]+)([^ ]+)([ ]+)([^ ]+)")
-	// lost with this regex ? http://regexr.com/3fe0m
-
 	d.Phys = make([]string, 0, numServers)
 	d.Virt = make([]string, 0, numServers)
 
 	for i := range interestingLines {
-		matches := regex.FindStringSubmatch(interestingLines[i])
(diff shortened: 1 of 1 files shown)


simul/platform/deterlab.go
```diff
@@ -345,17 +346,70 @@ func (d *Deterlab) Wait() error {
...
+	// lost with this regex ? http://regexr.com/3fe0m
+
 	d.Phys = make([]string, 0, numServers)
 	d.Virt = make([]string, 0, numServers)
-	for i := 1; i <= numServers; i++ {
-		d.Phys = append(d.Phys, fmt.Sprintf("server-%d.%s.%s", i-1, d.Experiment, name))
-		d.Virt = append(d.Virt, fmt.Sprintf("%s%d", ip, i))
+
+	for i := range interestingLines {
+		matches := regex.FindStringSubmatch(interestingLines[i])
```
ineiti: I discovered this method: `strings.Fields` - it does the same: https://play.golang.org/p/vOHXd80mji without regexp-magick (not that I don't like regexps, but...)

simul/platform/deterlab.go
```diff
@@ -345,17 +346,70 @@ func (d *Deterlab) Wait() error {
...
+
+		if len(matches) < 6 {
```
ineiti: Shouldn't this be `!=` instead of `<`?

lbarman: No, the API yields other infos (more than 6). This is just a check for when the API will change in background.
We can remove this weird check and pray that the API does not change, or keep it and maybe be warned. What do you think ?

ineiti: I think we need to rewrite the parser anyway and add some `_test`-method ;)

ineiti: the part of `createHosts` that parses the output.

ineiti:
Closing in favor of #141

PR #141, opened 2017-02-28
Deterlab nodes 140

### LINKED ISSUE #140: Deterlab node-detection
Now: the simulation on deterlab uses hardcoded values for server names, of the form
server-XXX.EXPERIMENT.PROJECT.isi.deterlab.net
10.255.0.X
Problem: Depending on the topology, this is really inconvenient. For two LANs for instance, we need different subsets of IPs, and server names are useful to differentiates machines.
Wanted: Have the deterlab-simulation auto-detect the nodes and their IPs.

ineiti:
Closes #140

### DIFF
--- a/simul/platform/deterlab.go
+++ b/simul/platform/deterlab.go
@@ -34,6 +34,8 @@ import (
 
 	"path/filepath"
 
+	"errors"
+
 	"github.com/BurntSushi/toml"
 	"github.com/dedis/onet"
 	"github.com/dedis/onet/log"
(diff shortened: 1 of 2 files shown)


simul/platform/deterlab.go
```diff
@@ -343,19 +343,53 @@ func (d *Deterlab) Wait() error {
...
+	apiReply, err := SSHRun(d.Login, d.Host, command)
+	if err != nil {
+		log.Fatal("Error while querying Deterlab:", err)
+	}
+	d.parseHosts(string(apiReply))
+}
+
+func (d *Deterlab) parseHosts(str string) {
+	// Get the link-information, which is the second block in `expinfo`-output
+	linkInfo := strings.Split(str, "\n\n")[1]
```
nikkolasg: Just so we look good programmer: check if the length is at least 2.

simul/platform/deterlab.go
```diff
@@ -343,19 +343,53 @@ func (d *Deterlab) Wait() error {
...
+	// Test for correct version in case the API-output changes
+	if !strings.HasPrefix(linkInfo, "Virtual Lan/Link Info:") {
+		log.Fatal("Didn't recognize output of 'expinfo'")
+	}
+	nodes := strings.Split(linkInfo, "\n")[3:]
```
nikkolasg: Just so we look good programmer: check if the length is at least 4.

ineiti:
The test passes, so it should work ;) I asked @lbarman to test it using his code, but he has to re-write part of it.

PR #143, opened 2017-03-06
added startup-checks

christophebertrand:
closes #824

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -19,7 +19,8 @@ import (
 	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
 
-	"github.com/cloudfoundry/gosigar"
+	"github.com/shirou/gopsutil/mem"
+
 	// CoSi-protocol is not part of the cothority.
 
 	// For the moment, the server only serves CoSi requests
@@ -311,8 +312,10 @@ func tryConnect(ip, binding network.Address) error {
 	<-listening
 	conn, err := net.Dial("tcp", ip.NetworkAddress())
 	if err != nil {
-		log.Error("Could not connect itself to public address:", err)
 	} else {
+		log.Panic("Could not connect itself to public address. This is most"+
+			" probably an error in your system-setup. Please make sure this conode "+
+			"can connect to", ip.NetworkAddress())
 		log.Info("Succesfully connected istelf to port")
 		conn.Close()
 	}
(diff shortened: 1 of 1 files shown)


app/server.go
```diff
@@ -291,20 +293,29 @@ func askReachableAddress(port string) network.Address {
...
+		listening <- true
 		con, _ := ln.Accept()
 		<-stopCh
 		con.Close()
 	}()
 	defer func() { stopCh <- true }()
+	<-listening
+	conn, err := net.Dial("tcp", ip.NetworkAddress())
+	if err != nil {
+		log.Error("Could not connect itself to public address:", err)
```
ineiti: I think this should `Panic` and not just `Error`. It's an error that needs to be corrected.
Add a hint for sysadmins:
"This is most probably an error in your system-setup. Please make sure this conode can connect to #{address}"

nikkolasg: But don't we have this problem for our own two conode ?

nikkolasg:
I have a concern about the usage of an external library which does not seem to be maintained no more (last commit 11months old, but usually are 2years old). It's something very specific and "simple". Also, we don't support windows, whereas cigar seems to support windows platform too. In that respect, would it be possible to take out the parts of sugar and make a small utility file doing the same thing ? 
Is it long ?

ineiti:
I didn't find anything on Cigar - is it golang?

nikkolasg:
Sorry I edited my comment, I mis-wrote the library. it's "gosigar". https://github.com/cloudfoundry/gosigar

ineiti:
Hmm - that also looks quite old to me. I'm OK with keeping the current, but if you think it's necessary to switch to `gosigar`. I'm happy with the current version. Does it work with go 1.8?

app/server.go
```diff
@@ -311,8 +312,10 @@ func tryConnect(ip, binding network.Address) error {
 	<-listening
 	conn, err := net.Dial("tcp", ip.NetworkAddress())
 	if err != nil {
-		log.Error("Could not connect itself to public address:", err)
 	} else {
+		log.Panic("Could not connect itself to public address. This is most"+
+			" probably an error in your system-setup. Please make sure this conode "+
+			"can connect to", ip.NetworkAddress())
```
ineiti: That's a strange `if`-branch... `Panic`, then `Info`, then `Close`. Best you do:
```
conn, err :=
log.ErrFatal("Could not...")
log.Info()
conn.Close()
```

app/server.go
```diff
@@ -291,20 +294,31 @@ func askReachableAddress(port string) network.Address {
...
 		}
+		listening <- true
 		con, _ := ln.Accept()
 		<-stopCh
 		con.Close()
 	}()
 	defer func() { stopCh <- true }()
+	<-listening
+	conn, err := net.Dial("tcp", ip.NetworkAddress())
+	if err != nil {
```
ineiti: That's wrong, sorry. Try to follow the error-path if err == nil and if err != nil. I proposed the following:

```
log.ErrFatal(err, "Could not connect itself to public address. This is most "+
			"probably an error in your system-setup. Please make sure this conode "+
			"can connect to", ip.NetworkAddress())

log.Info("Succesfully connected istelf to port")
conn.Close()
```

ineiti:
I wanted to create a new branch with github, but it failed twice on me, so I did it directly to your branch - sorry.

Issue #116, opened 2017-02-07
Rename in `crypto/key.go`


ineiti:
If we keep this file, we should rename all functions to be more uniform. At least replace `Pub` with `Point`. But probably this should go into `crypto.v1` anyway, and then this can be removed.

PR #144, opened 2017-03-07
Rename every 'Public' in abstract 'Point' instead

AlexConnat:
A lot of these functions contains a abstract.Suite as first argument, which is unused. Why is so?
Solved #116

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -53,7 +53,7 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	point, err := crypto.StringHexToPub(network.Suite, hc.Public)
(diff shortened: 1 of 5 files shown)


ineiti:
Something we can't port to v1 I think, as it will break some of the code in `cothority` and others. Else it's OK for me.

nikkolasg:
@AlexConnat I don't see any unused suite arguments in the changes you made at least... Where did you find that ?
Else it's OK.

ineiti:
@nikkolasg I think he meant in general in `keys.go`:
```
WriteHexPub
WriteHexScalar
PubToStringHex
PubToString64
ScalarToStringHex
ScalarToString64
```

I think I wanted to keep the `suite` in there for it to be similar to the other functions. But now it seems not useful and more logical to take `suite` away. What do you think?

PR #145, opened 2017-03-09
starting simulation.Node only once all conodes are started

ineiti:
In simulation-mode, this adds a listener to all servers, so that it can receive a broadcast-message and then start the `simulation.Node`. This makes sure that all nodes are up before `simulation.Node` is called.

### DIFF
--- a/simul/platform/runsimul.go
+++ b/simul/platform/runsimul.go
@@ -62,18 +62,19 @@ func Simulate(serverAddress, simul, monitorAddress string) error {
 			return err
 		}
 		sims[i] = sim
-		func(sc *onet.SimulationConfig) {
-			server.RegisterProcessorFunc(simulInitID, func(env *network.Envelope) {
-				err = sim.Node(sc)
-				if err != nil {
-					log.Error(err)
-				}
-				sc.Server.Send(env.ServerIdentity, &simulInitDone{})
-			})
-			server.RegisterProcessorFunc(simulInitDoneID, func(env *network.Envelope) {
-				wgSimulInit.Done()
-			})
-		}(sc)
+		// Need to store sc in a tmp-variable so it's correctly passed
+		// to the Register-functions.
+		scTmp := sc
+		server.RegisterProcessorFunc(simulInitID, func(env *network.Envelope) {
+			err = sim.Node(scTmp)
+			if err != nil {
+				log.Error(err)
+			}
+			scTmp.Server.Send(env.ServerIdentity, &simulInitDone{})
+		})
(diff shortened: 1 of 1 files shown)


simul/platform/runsimul.go
```diff
@@ -99,6 +105,10 @@ func Simulate(serverAddress, simul, monitorAddress string) error {
 			timeout *= 2
 		}
 		childrenWait.Record()
+		log.Lvl2("Broadcasting start")
+		for _, conode := range rootSC.Tree.Roster.List {
+			rootSC.Server.Send(conode, &startSimulation{})
+		}
```
nikkolasg: Should we not wait that all `Node` methods have been executed first before going on ?

ineiti: Good question - I don't think it's necessary. This is the root-node that contacts all other nodes (including itself). What happens if one conode is still in `simulation.Node` and then the root-node calls its `simulation.Root`? Is it bad?

nikkolasg: As long as we're in it, I'd say it's better to do it, because `Node` is used as a `Init` function. If the init is not finished, it seems less useful than it can be because we're going back to point 0: the users have to code a synchronization mechanism themself.

ineiti:
I tested this on Deterlab with 3'000 nodes on 10 servers for the count-protocol.

simul/platform/runsimul.go
```diff
@@ -55,11 +61,19 @@ func Simulate(serverAddress, simul, monitorAddress string) error {
...
+				err = sim.Node(sc)
+				if err != nil {
+					log.Error(err)
+				}
+				sc.Server.Send(env.ServerIdentity, &simulInitDone{})
+			})
+			server.RegisterProcessorFunc(simulInitDoneID, func(env *network.Envelope) {
+				wgSimulInit.Done()
+			})
+		}(sc)
```
nikkolasg: Why do you put that in a func ? You don't call it in a goroutine, neither sc is changing... ?

ineiti: I did some tests in go-playground which seemed to show that I cannot rely on `sc` being correctly passed to the `ProcessorFunc`, except if I pass it as an argument to a function.

nikkolasg: Mhhh I see. If you simply create a variable to store the current value of sc **inside the loop**, it should work without the `func()`
```
for sc .... {
   mySc := sc
   RegisterProcessor(mySc)
}
```
The tmp variable is ony available from the current iteration so should not change. At least it does not in [go playground](https://play.golang.org/p/BwsqXYcJ81)

PR #150, opened 2017-03-17
added migration script

christophebertrand:
closes #821


ineiti:
Cool - can you please add:

- verification we're not in the `onet`-directory
- a warning that it changes the files and a `read`-`y`-verification? Something like http://stackoverflow.com/questions/226703/how-do-i-prompt-for-yes-no-cancel-input-in-a-linux-shell-script

ineiti:
Very nice

PR #153, opened 2017-03-22
modified definition of service functions

### LINKED ISSUE #146: Change definition of service-functions
Now: `ServiceProcessor.RegisterHandler` wants to have `network.Message` as first return-argument.

Wanted: accept any message-type as first return-argument

Further work: re-write services so that the return-type is the _actual_ type of the message.

Why: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

christophebertrand:
closes #146


nikkolasg:
Looking good to me !

ineiti:
Yes, +1. Even with tests!

PR #155, opened 2017-03-22
fixed toml parsing

### LINKED ISSUE #152: Simulation silently hangs on valid toml
If all lines are of the form
```
key1=val1
```
the simulation hangs silently. 
To fix, the last line must be of the form
```
keyY=valY

keyZ
valZ
```

christophebertrand:
closes #152

### DIFF
--- a/simul/platform/platform.go
+++ b/simul/platform/platform.go
@@ -119,8 +119,7 @@ func ReadRunFile(p Platform, filename string) []*RunConfig {
 		log.Lvlf5("Platform is now %+v", p)
 	}
 
-	for {
-		scanner.Scan()
+	for scanner.Scan() {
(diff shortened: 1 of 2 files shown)


nikkolasg:
A test would be good :)

ineiti:
Thinking out loud: shouldn't it fail if we don't have the tail with

```
key1[, key2]
value1[, value2]
```
? Usually that tail indicates the experiments we want to run. So if the tail is missing, it's like it's just 1 experiment. Isn't this confusing?

nikkolasg:
@ineiti I'm ok with **not** failing if there is no "csv" part since one can run one experiment...

nikkolasg:
LGTM for me.

PR #160, opened 2017-03-24
replaced Fatal with Lvl2 and added test

### LINKED ISSUE #159: `GenerateNaryTreeWithRoot` should not `Fatal`
Now: If the given root in `GenerateNaryTreeWithRoot` doesn't exist, it `Fatal`s.

Wanted: return `nil` if something goes wrong.

Why not return `(*Tree, error)`? - because this would break API-compatibility with `v1`.

ineiti:
Closes #159

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -474,7 +474,7 @@ func (el *Roster) GenerateBigNaryTree(N, nodes int) *Tree {
 func (el *Roster) GenerateNaryTreeWithRoot(N int, rootServerIdentity *network.ServerIdentity) *Tree {
 	rootIndex, _ := el.Search(rootServerIdentity.ID)
 	if rootIndex < 0 {
-		log.Fatal(rootServerIdentity, el.List, log.Stack())
+		log.Lvl2("Asked for non-existing root:", rootServerIdentity, el.List, log.Stack())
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -474,7 +474,7 @@ func (el *Roster) GenerateBigNaryTree(N, nodes int) *Tree {
 func (el *Roster) GenerateNaryTreeWithRoot(N int, rootServerIdentity *network.ServerIdentity) *Tree {
 	rootIndex, _ := el.Search(rootServerIdentity.ID)
 	if rootIndex < 0 {
-		log.Fatal(rootServerIdentity, el.List, log.Stack())
+		log.Lvl2("Asked for non-existing root:", rootServerIdentity, el.List, log.Stack())
```
nikkolasg: I'm Ok with not panicking but then you must return an error, don't you think ?

nikkolasg: and the `log.Stack` has no meaning anymore.

ineiti: I'll remove `log.Stack`, but won't add an error, as this would break `v1`-API-compatibility (as explained in the issue #159). For the moment it's: return `nil` -> it failed.

PR #167, opened 2017-03-31
adding mutex for testing

### LINKED ISSUE #165: race-condition in Context.Save
`Context.Save` accesses the _global_ `contextData`-variable while in testing-mode. And `contextData` is not protected by a mutex.

ineiti:
Closes #165

### DIFF
--- a/context.go
+++ b/context.go
@@ -126,8 +126,10 @@ func (c *Context) String() string {
 	return c.server.ServerIdentity.String()
 }
 
-var contextData = map[string][]byte{}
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -127,6 +127,7 @@ func (c *Context) String() string {
 }
 
 var contextData = map[string][]byte{}
```
nikkolasg: can rename that to `testContextData` or stg ? And also the lock?

ineiti:
struct or not to struct - that is the question...

PR #170, opened 2017-04-07
adding correct buffer to receive

ineiti:
Correct measuring of receiving byes.


nikkolasg:
do we have a test for making sure the Rx() and Tx() methods behave correctly ?

ineiti:
Updated test for `Tx` and `Rx`. It fails with the old code and passes with the correction.

PR #172, opened 2017-04-13
adding flag to print every round-stat instead of averageing them

### LINKED ISSUE #171: Return more than one line in an experiment
Now: each experiment-run (one line in `experiment.toml`) in `onet/simul` results in one line in the `experiment.csv`-file.

Wanted: being able to return more than one line to the `experiment.csv`-file.

Why: e.g. latest experiments for the paper has a high setup-time, and I'd like to run multiple measurements on the same setup.

ineiti:
Closes #171 

Also includes simplification of the simulation-code that has never been used: re-starting failing simulations.

### DIFF
--- a/simul/build.go
+++ b/simul/build.go
@@ -155,8 +155,9 @@ func RunTests(name string, runconfigs []*platform.RunConfig) {
 		if i == 0 {
 			stats.WriteHeader(f)
 		}
-		if strings.ToLower(rc.Get("EveryRoundStats")) == "true" {
-			stats.WriteValuesEveryRoundStats(f)
+		if rc.Get("IndividualStats") != "" {
+			err := stats.WriteIndividualStats(f)
+			log.ErrFatal(err)
 		} else {
 			stats.WriteValues(f)
 		}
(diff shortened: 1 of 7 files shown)


ineiti:
gives error when running in a multi-node system because of the `bandwidth_rx` in the `simul/platform/runsimul::simulate`...

simul/build.go
```diff
@@ -150,31 +146,20 @@ func RunTests(name string, runconfigs []*platform.RunConfig) {
...
+			log.Error("Error running test, trying again:", err)
 			continue
 		}
 
-		s := monitor.AverageStats(runs)
 		if i == 0 {
-			s.WriteHeader(f)
+			stats.WriteHeader(f)
+		}
+		if strings.ToLower(rc.Get("EveryRoundStats")) == "true" {
```
nikkolasg: Can we call it more like `IndividualMeasures` (to be an antonym of `aggregate measures` ? 
And just `if rc.Get("IndividualMeasures") != "" { ... }` should be enough ?

ineiti: OK - sounds good.

PR #177, opened 2017-04-28
passes config to all children

### LINKED ISSUE #147: config-message passing error
Now: if we use `TreeNodeInstance.SetConfig` to send a config for a new protocol, it is only sent to one node.

Problem: if a node has two or more children, only one child gets the config. This is bad

Wanted: each node keeps a list in `TreeNodeInstance.SendTo` of nodes that already received the config. Only if it's the first message will the config be sent along.

christophebertrand:
closes #147

### DIFF
--- a/treenode.go
+++ b/treenode.go
@@ -52,6 +52,7 @@ type TreeNodeInstance struct {
 	// config is to be passed down in the first message of what the protocol is
 	// sending if it is non nil. Set with `tni.SetConfig()`.
 	config    *GenericConfig
+	sentTo    map[TreeNodeID]bool
 	configMut sync.Mutex
 }
 
@@ -77,6 +78,7 @@ func newTreeNodeInstance(o *Overlay, tok *Token, tn *TreeNode, io MessageProxy) 
 		msgDispatchQueue:     make([]*ProtocolMsg, 0, 1),
 		msgDispatchQueueWait: make(chan bool, 1),
 		protoIO:              io,
+		sentTo:               make(map[TreeNodeID]bool),
 	}
 	go n.dispatchMsgReader()
 	return n
@@ -126,9 +128,9 @@ func (n *TreeNodeInstance) SendTo(to *TreeNode, msg interface{}) error {
 	var c *GenericConfig
 	// only sends the config once
 	n.configMut.Lock()
-	if n.config != nil {
+	if !n.sentTo[to.ID] {
 		c = n.config
-		n.config = nil
+		n.sentTo[to.ID] = true
 	}
 	n.configMut.Unlock()
 
@@ -679,6 +681,9 @@ func (n *TreeNodeInstance) SetConfig(c *GenericConfig) {
 	n.configMut.Lock()
 	defer n.configMut.Unlock()
 	n.config = c
+	for id := range n.sentTo {
+		n.sentTo[id] = false
+	}
 }
 
 func (n *TreeNodeInstance) isBound() bool {
--- a/treenode.go
+++ b/treenode.go
@@ -677,13 +677,14 @@ func (n *TreeNodeInstance) TreeNodeInstance() *TreeNodeInstance {
 // SetConfig sets the GenericConfig c to be passed down in the first message
 // alongside with the protocol if it is non nil. This config can later be read
 // by Services in the NewProtocol method.
-func (n *TreeNodeInstance) SetConfig(c *GenericConfig) {
+func (n *TreeNodeInstance) SetConfig(c *GenericConfig) error {
 	n.configMut.Lock()
 	defer n.configMut.Unlock()
-	n.config = c
-	for id := range n.sentTo {
-		n.sentTo[id] = false
+	if n.config != nil {
+		return errors.New("Can't set config twice")
 	}
+	n.config = c
+	return nil
 }
 
 func (n *TreeNodeInstance) isBound() bool {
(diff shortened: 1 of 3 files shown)


treenode.go
```diff
@@ -679,6 +681,9 @@ func (n *TreeNodeInstance) SetConfig(c *GenericConfig) {
 	n.configMut.Lock()
 	defer n.configMut.Unlock()
 	n.config = c
+	for id := range n.sentTo {
+		n.sentTo[id] = false
```
nikkolasg: The default value for a boolean is already "false" so I think this initialization step can be avoided.

christophebertrand: does `SetConfig` get called only once? If not one needs to make sure to clean the map so the new config gets send to everyone. Otherwise I agree that it is not needed

ineiti: I think it should be called only once. One could even argue if we want to enforce that.

treenode_test.go
```diff
@@ -41,6 +42,54 @@ func TestHandlerReturn(t *testing.T) {
...
+			ConfigMsgID)
+	}
+
+	network.RegisterMessage(dummyMsg{})
+	rootInstance, _ := local.NewTreeNodeInstance(tree.Root, spawnName)
+	rootInstance.SetConfig(&GenericConfig{serviceConfig})
+	err = rootInstance.SendToChildren(&dummyMsg{})
+	log.ErrFatal(err)
+	// wait until the processor has processed the expected number of config messages
+	<-done
```
ineiti: You should do a `select` on the channel and add a timeout:

```
select{
case <-done:
case <-time.After(time.Second):
    t.Fatal("Didn't receive response in time")
}
```

treenode.go
```diff
@@ -126,9 +128,9 @@ func (n *TreeNodeInstance) SendTo(to *TreeNode, msg interface{}) error {
 	var c *GenericConfig
 	// only sends the config once
 	n.configMut.Lock()
-	if n.config != nil {
+	if !n.sentTo[to.ID] {
 		c = n.config
-		n.config = nil
+		n.sentTo[to.ID] = true
 	}
```
ineiti: Should we add a `n.config = nil` once the configuration has been sent to all nodes? This would help the garbage-collector.

On the other hand, the protocol is supposed to finish _some_ time, then it will be cleaned up a bit later.

nikkolasg: And now your question has made me think about other issues... So usually, we *assume* the first message MUST contains the configuration. But is that really what we want all the time ? It can probably happen that a config has to be sent to all *children* and then in case stuff goes wrong, the root contact the sub-children (for example, what our theoretical CoSi would likely to if it were to exist). The message for the sub-children should *not* contain the config. 

In any way, this specific case is already handled at the recipient level, where the Overlay will only look for the config if the TreeNode already exists, but that means the config received from the root (to the subchildren) will stay in memory forever... :/

For now let's move on with this approach but we'll have to dig more afterwards.  @christophebertrand if you think of stg, don't hesitate to chim in ;)

ineiti: The first message CAN contain the message, was that what you meant?

And I don't see why in your example the message for the grand-children should not contain the configuration.

nikkolasg: Because the grand children are already "alive",i.e. created. THe configuration can only be passed during the *creation* of a `ProtocolInstance` in a `Service`, not during its lifetime.

treenode.go
nikkolasg: Indeed, it should only be called once. So I'd still remove this initialization, and either a) a big warning in the comments to only call this method once, or b) check in `SetConfig` if it has been called more than once, then panic.
I'm more for the a) approach.

christophebertrand:
I am wondering if currently all the nodes will get the config. If the nodes are only communicating with their direct relatives (parents, children), the config message will never get passed down to the grand-children. Or am I missing something?

nikkolasg:
This is done at the `Service` level, where the `Service` will instantiate the protocol and also set the config, so the grand children will also get the config etc etc.

treenode.go
ineiti: It returns an `error`, so it can be `Can't send config twice`.

treenode.go
ineiti: True. So we can only wait for the protocolinstance to be closed and then free the config.

ineiti:
@nikkolasg - so you mean if the root-node sets a configuration, the intermediate-nodes don't need to set it again? Or do you expect the service of the intermediate-nodes to set the configuration again?

nikkolasg:
@ineiti good question. I expect the service of the intermediate nodes to set it again, as it might change. But it's debatable if we should do it automatically, although that would require us to allow more than one time to `SetConfig` since a `Service` might want to put a different `Config` that the one the root node put initially. 
So I'm still more in favor of the explicit way: every `Service` should set the `Config` in its `ProtocolInstance`.

treenode_test.go
```diff
@@ -41,6 +43,58 @@ func TestHandlerReturn(t *testing.T) {
...
+			RequestTreeMsgID,
+			SendTreeMsgID,
+			RequestRosterMsgID,
+			SendRosterMsgID,
+			ConfigMsgID)
+	}
+
+	network.RegisterMessage(dummyMsg{})
+	rootInstance, _ := local.NewTreeNodeInstance(tree.Root, spawnName)
+	rootInstance.SetConfig(&GenericConfig{serviceConfig})
```
ineiti: Needs to verify `err == nil`.
Add a second line with `SetConfig` and verify `err != nil`.

PR #178, opened 2017-04-28
made ID use consistent - 1/2

christophebertrand:


ineiti:
Linting errors - I changed the title to have "- 1/2" in it.

messages.go
```diff
@@ -58,12 +58,33 @@ func (rId RoundID) String() string {
 	return uuid.UUID(rId).String()
 }
 
+// Equal returns true if and only if the given RoundID equals the current one.
+func (rId RoundID) Equal(rId2 RoundID) bool {
+	return uuid.Equal(uuid.UUID(rId), uuid.UUID(rId2))
+}
+
+//IsNil returns true iff the RoundID is Nil
```
ineiti: `s://Is:// Is:`

messages.go
```diff
@@ -58,12 +58,33 @@ func (rId RoundID) String() string {
...
+func (t TokenID) String() string {
+	return uuid.UUID(t).String()
+}
+
+// Equal returns true if and only if the given TokenID equals the current one.
+func (t TokenID) Equal(t2 TokenID) bool {
+	return uuid.Equal(uuid.UUID(t), uuid.UUID(t2))
+}
+
+//IsNil returns true iff the TokenID is Nil
```
ineiti: `s://Is:// Is:`

network/encoding.go
```diff
@@ -49,6 +49,11 @@ func (mId MessageTypeID) Equal(t MessageTypeID) bool {
 	return bytes.Compare(uuid.UUID(mId).Bytes(), uuid.UUID(t).Bytes()) == 0
 }
 
+//IsNil returns true iff the MessageTypeID is Nil
```
ineiti: `s://Is:// Is:`

network/struct.go
```diff
@@ -83,9 +83,19 @@ type ServerIdentity struct {
...
+}
+
 // Equal returns true if both ServerIdentityID are equal or false otherwise.
-func (eid ServerIdentityID) Equal(other ServerIdentityID) bool {
-	return uuid.Equal(uuid.UUID(eid), uuid.UUID(other))
+func (eId ServerIdentityID) Equal(other ServerIdentityID) bool {
+	return uuid.Equal(uuid.UUID(eId), uuid.UUID(other))
+}
+
+//IsNil returns true iff the ServerIdentityID is Nil
```
ineiti: CP:
`s://Is:// Is:`

protocol.go
```diff
@@ -17,6 +17,16 @@ func (pid ProtocolID) String() string {
 	return uuid.UUID(pid).String()
 }
 
+// Equal returns true if and only if the given ProtocolID equals the current one.
+func (pid ProtocolID) Equal(pid2 ProtocolID) bool {
+	return uuid.Equal(uuid.UUID(pid), uuid.UUID(pid2))
+}
+
+//IsNil returns true iff the ProtocolID is Nil
```
ineiti: CP:
`s://Is:// Is:`

tree.go
```diff
@@ -49,6 +49,11 @@ func (tId TreeID) String() string {
 	return uuid.UUID(tId).String()
 }
 
+//IsNil returns true iff the TreeID is Nil
```
ineiti: CP:
`s://Is:// Is`

tree.go
```diff
@@ -349,6 +354,16 @@ func (elId RosterID) String() string {
 	return uuid.UUID(elId).String()
 }
 
+// Equal returns true if and only if the given RosterID equals the current one.
+func (elId RosterID) Equal(elId2 RosterID) bool {
+	return uuid.Equal(uuid.UUID(elId), uuid.UUID(elId2))
+}
+
+//IsNil returns true iff the RosterID is Nil
```
ineiti: CopyPaste

tree.go
```diff
@@ -559,6 +574,11 @@ func (tId TreeNodeID) Equal(tID2 TreeNodeID) bool {
 	return uuid.Equal(uuid.UUID(tId), uuid.UUID(tID2))
 }
 
+//IsNil returns true iff the TreeNodID is Nil
```
ineiti: CopyPaste

service.go
```diff
@@ -46,8 +46,18 @@ type Service interface {
 type ServiceID uuid.UUID
 
 // String returns the string representation of this ServiceID
-func (s *ServiceID) String() string {
-	return uuid.UUID(*s).String()
+func (s ServiceID) String() string {
+	return uuid.UUID(s).String()
+}
+
+// Equal returns true if and only if the given ServiceID equals the current one.
```
nikkolasg: `... equals s`

tree.go
```diff
@@ -39,8 +39,8 @@ type Tree struct {
 // TreeID uniquely identifies a Tree struct in the onet framework.
 type TreeID uuid.UUID
 
-// Equals returns true if and only if the given TreeID equals the current one.
-func (tId TreeID) Equals(tID2 TreeID) bool {
+// Equal returns true if and only if the given TreeID equals the current one.
```
nikkolasg: ` ... equals tId`

tree.go
```diff
@@ -349,6 +354,16 @@ func (elId RosterID) String() string {
 	return uuid.UUID(elId).String()
 }
 
+// Equal returns true if and only if the given RosterID equals the current one.
```
nikkolasg: ` ... equals elId`

service.go
ineiti: I think by `given ServiceID` he means `s`. So it could be:
```
// Equal returns true if and only if `s` equals the current one.
```

tree.go
ineiti: as above

tree.go
ineiti: as above

messages.go
```diff
@@ -58,12 +58,33 @@ func (rId RoundID) String() string {
 	return uuid.UUID(rId).String()
 }
 
+// Equal returns true if and only if the given RoundID equals rId.
```
ineiti: I think we asked you to mis-correct this one: we have an argument, `rID2`, and a method on a structure `RoundID`. So I think the correct comment would be:

```
// Equal returns true if rID2 equals this RoundID
```

and also for the other `Equals`. This way we also describe the argument of the method instead of referencing it only as `given RoundID`.

PR #179, opened 2017-05-09
Added validHostname as isolated function.

marcozov:
Added few (temporary) tests
It has to be integrated with the rest

### DIFF
--- a/network/address.go
+++ b/network/address.go
@@ -61,37 +61,66 @@ func (a Address) ConnType() ConnType {
 	return connType(vals[0])
 }
 
+// IsHostname returns true if the address is defined by a VALID DNS name
+func (a Address) IsHostname() bool {
+	host := a.Host()
+
+	// validHostname(host) would be enough with the current implementation.
+	// However, if we will include IDNs as valid hostnames, an IP address in the form
+	// *.*.*.* would be a valid hostname too. This is why ParseIP is used as well.
+	return validHostname(host) && net.ParseIP(host) == nil
+}
+
 // NetworkAddress returns the network address part of the address, which is
-// the IP address and the port joined by a colon.
-// If the address is defined with a hostname, a lookup is performed and (one of)
-// the related IP address is returned
-// It returns an empty string if the a.Valid() returns false.
+// the host and the port joined by a colon.
+// It returns an empty string the address is not valid
 func (a Address) NetworkAddress() string {
 	if !a.Valid() {
 		return ""
 	}
 	vals := strings.Split(string(a), typeAddressSep)
-	validNetworkAddress := vals[1]
+	return vals[1]
+}
 
-	host, port, err := net.SplitHostPort(validNetworkAddress)
-	if err != nil {
+// NetworkAddressResolved returns the network address of the address, but resolved.
+// That is: the hostname resolved and the port joined by a colon.
+// It returns an empty string if the address is not valid.
+func (a Address) NetworkAddressResolved() string {
+	if !a.Valid() {
+		return ""
+	}
+	ipAddress := a.Resolve()
+	port := a.Port()
+	return net.JoinHostPort(ipAddress, port)
+}
+
+// Resolve returns the IP address associated to the hostname that represents the address a.
+// If a is defined by an IP address (*.*.*.*) or if the hostname is not valid, the empty string
+// is returned
+func (a Address) Resolve() string {
+	if !a.Valid() {
+		return ""
+	}
+	host := a.Host()
+	// Ipv6 not handled properly yet
+	if host == "[::]" {
+		return "::"
+	}
+	// If the address is defined by an IP address, return it
+	if net.ParseIP(host) != nil {
+		return host
+	}
+
+	if !a.IsHostname() {
 		return ""
 	}
 
-	// checking the validity of the IP address (both IPv4 or IPv6)
-	if net.ParseIP(host) != nil {
-		return validNetworkAddress
-	}
-
-	// if host is not a valid IP address (i.e. ParseIP returns nil),
-	// it must be a hostname (since at the beginning we check whether the address is valid
 	ipAddress, err := lookupHost(host)
 	if err != nil {
 		return ""
 	}
 
-	// it returns the first found ip address
-	return net.JoinHostPort(ipAddress[0], port)
+	return ipAddress[0]
 }
 
 // validHostname returns true if the hostname is well formed or false otherwise.
@@ -215,7 +244,7 @@ func (a Address) Public() bool {
 	private, err := regexp.MatchString("(^127\\.)|(^10\\.)|"+
 		"(^172\\.1[6-9]\\.)|(^172\\.2[0-9]\\.)|"+
 		"(^172\\.3[0-1]\\.)|(^192\\.168\\.)|(^169\\.254)|"+
-		"(^\\[::\\])", a.NetworkAddress())
+		"(^\\[::\\])", a.NetworkAddressResolved())
 	if err != nil {
 		return false
 	}
--- a/network/address.go
+++ b/network/address.go
@@ -3,7 +3,6 @@ package network
 import (
 	"net"
 	"regexp"
-	"sort"
 	"strconv"
 	"strings"
 )
@@ -18,6 +17,8 @@ type ConnType string
 // by a colon.
 // It doesn't support IPv6 yet.
 type Address string
+
+var lookupHost func(string) ([]string, error) = net.LookupHost
 
 const (
 	// PlainTCP is an unencrypted TCP connection.
@@ -62,6 +63,8 @@ func (a Address) ConnType() ConnType {
 
 // NetworkAddress returns the network address part of the address, which is
 // the IP address and the port joined by a colon.
+// If the address is defined with a hostname, a lookup is performed and (one of)
+// the related IP address is returned
 // It returns an empty string if the a.Valid() returns false.
 func (a Address) NetworkAddress() string {
 	if !a.Valid() {
@@ -82,12 +85,11 @@ func (a Address) NetworkAddress() string {
 
 	// if host is not a valid IP address (i.e. ParseIP returns nil),
 	// it must be a hostname (since at the beginning we check whether the address is valid
-	ipAddress, err := net.LookupHost(host)
+	ipAddress, err := lookupHost(host)
 	if err != nil {
 		return ""
 	}
 
-	sort.Strings(ipAddress)
 	// it returns the first found ip address
 	return net.JoinHostPort(ipAddress[0], port)
 }
@@ -102,22 +104,8 @@ func (a Address) NetworkAddress() string {
 //	- labels cannot start with a hyphen
 //	- labels cannot end with a hyphen
 //	- the last label is alphabetic
-// sources about the syntax of hostnames:
-//	- https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names
-//	- https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#IDN_Test_TLDs
-//	- RFC 1035, RFC 1123
-//	Theoretically, it seems that a TLD with a hyphen (-) is accepted:
-// 		-> no RFC says the opposite, explicitly
-//		-> The second source shows some examples of TLD with hyphens (the ones used for punycode)
-//	But: https://tools.ietf.org/html/rfc1123#page-13 has the following lines
-//		-> "However, a valid host name can never
-//		have the dotted-decimal form #.#.#.#, since at least the
-//		highest-level component label will be alphabetic."
-//
-
-//	I guess it depends on whether we want to handle IDNs too (whose representation was
-// 	introduced in a second moment, in RFC 3492).
-//	At moment, they are NOT handled. A minor change would be enough to handle them.
+//	More information about the definition of the TLD (the last label of a hostname) on:
+//		https://github.com/dedis/cothority/issues/620
 // This method assumes that only the host part is passed as parameter
 // This function is integrated in the Valid() function
 func validHostname(s string) bool {
--- a/network/address.go
+++ b/network/address.go
@@ -67,7 +67,27 @@ func (a Address) NetworkAddress() string {
 		return ""
 	}
 	vals := strings.Split(string(a), typeAddressSep)
-	return vals[1]
+	validNetworkAddress := vals[1]
+
+	host, port, err := net.SplitHostPort(validNetworkAddress)
+	if err != nil {
+		return ""
+	}
+
+	// checking the validity of the IP address (both IPv4 or IPv6)
+	if net.ParseIP(host) != nil {
+		return validNetworkAddress
+	}
+
+	// if host is not a valid IP address (i.e. ParseIP returns nil),
+	// it must be a hostname (since at the beginning we check whether the address is valid
+	ipAddress, err := net.LookupHost(host)
+	if err != nil {
+		return ""
+	}
+
+	// it returns the first found ip address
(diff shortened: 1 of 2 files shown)


network/address.go
```diff
@@ -70,6 +70,50 @@ func (a Address) NetworkAddress() string {
...
+
+	labels := strings.Split(s, ".")
+
+	for _, element := range labels {
+		if len(element) < 1 || len(element) > 63 {
+			return false
+		}
+	}
+
+	valid, _ := regexp.MatchString("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)+([A-Za-z]|[A-Za-z][A-Za-z\\-]*[A-Za-z])$", s)
```
ineiti: If I read the regular expression correctly, this implies that the TLD could have a `-` inside - is that correct? As far as I remember, TLDs are only comprised of letters.

ineiti: to make it simpler, convert everything to lower-case and match against this - makes the regular expression more readable.

network/address.go
```diff
@@ -70,6 +70,50 @@ func (a Address) NetworkAddress() string {
...
+	return valid
+}
+
+//func connType(t string) ConnType
```
ineiti: where does that come from?

network/address.go
marcozov: By looking at some RFCs ( https://tools.ietf.org/html/rfc1123 , https://tools.ietf.org/html/rfc1035 ) it seems that nothing forbids TLD with hyphens, as long as it is not the starting/ending character (just like the other labels. 
Also, these kinds of TLDs are used when punycode is involved: https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#IDN_Test_TLDs .

marcozov: But it also seems like it is contradicting itself: according to what I have just posted, nothing prevents a TDL to also have numbers in it. But this part "However, a valid host name can never have the dotted-decimal form #.#.#.#, since at least the highest-level component label will be alphabetic" in RFC 1123 clearly states a different thing.
I think the simplest thing, at least for now, is to ignore these kinds of TLDs and therefore my regular expression will accept only TLDs that are purely alphabetic. In case, I can easily adapt it later.

nikkolasg:
@marcozov I see in your last commit :
> A refactor of the Valid() function might be needed (for instance: the check ip == "localhost" is already handled by validHostname

If you feel you have a better way, please go ahead. Maybe do that in a separate PR first to see if we agree, but don't hesitate :)
Otherwise, good work so far !

marcozov:
@nikkolasg I just removed the line in which the check for localhost is performed: 
`
if ip == "localhost" {
   return true
} else if net.ParseIP(ip) == nil {
   return validHostname(ip)
}
return true
`
became:
`
	if net.ParseIP(ip) == nil {
		return validHostname(ip)
	}
return true
`
since "localhost" is handled inside the new function validHostname() (and net.ParseIP() does not handle it). I hope it is not a problem if I didn't do a new PR only for this thing, since it is strictly related to the issue of supporting the DNS.

network/address.go
```diff
@@ -67,7 +68,81 @@ func (a Address) NetworkAddress() string {
...
+	}
+
+	// if host is not a valid IP address (i.e. ParseIP returns nil),
+	// it must be a hostname (since at the beginning we check whether the address is valid
+	ipAddress, err := net.LookupHost(host)
+	if err != nil {
+		return ""
+	}
+
+	sort.Strings(ipAddress)
```
nikkolasg: Why would we need to sort it ?

network/address.go
```diff
@@ -67,7 +68,81 @@ func (a Address) NetworkAddress() string {
 		return ""
```
nikkolasg: Can you change the comments of the function, now that it performs a DNS Lookup ?

network/address.go
```diff
@@ -67,7 +68,81 @@ func (a Address) NetworkAddress() string {
...
+//	- labels have only ASCII letters 'a' through 'z' (case-insensitive), the digits
+//		'0' through '9', and the hyphen (-). No other symbol is permitted
+//	- labels cannot start with a hyphen
+//	- labels cannot end with a hyphen
+//	- the last label is alphabetic
+// sources about the syntax of hostnames:
+//	- https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names
+//	- https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#IDN_Test_TLDs
+//	- RFC 1035, RFC 1123
+//	Theoretically, it seems that a TLD with a hyphen (-) is accepted:
```
nikkolasg: All of the below discussion can be removed from the code. Specify as much as possible in the issue related to this PR and add a link to the issue here so we can reference it later.

network/address_test.go
```diff
@@ -27,6 +29,12 @@ func TestConnType(t *testing.T) {
 }
 
 func TestAddress(t *testing.T) {
+	localhostNetworkAddresses, _ := net.LookupHost("localhost")
+	sort.Strings(localhostNetworkAddresses)
+	facebookNetworkAddresses, _ := net.LookupHost("facebook.com")
+	sort.Strings(facebookNetworkAddresses)
+	googleNetworkAddresses, _ := net.LookupHost("google.com")
```
nikkolasg: Mhhhh... I'm not sure I like the idea of doing a DNS resolution for each test run:
 - It makes our whole test suite depending on the internet network connectivity
 - do we really want to fetch facebook and google addresses ?? I would have prefer some respectable institution like EPFL, EFF, and such... 

Usually, in these situations, I'd prefer to leverage a Golang interface:
```
type Resolver interface {
    LookupHost(string) ([]string,error)
}
```
One implementation would be just a proxy to `net.LookupHost` and another one would be a dummy implementation returning static (correct) values, previously fetched with a call to `net.LookupHost`.
The tests should use the latter.
What do you think ?

network/address_test.go
```diff
@@ -63,3 +80,22 @@ func TestAddress(t *testing.T) {
...
+	assert.True(t, validHostname("a.a"), "valid")
+	assert.True(t, validHostname("localhost"), "valid")
+	assert.True(t, validHostname("www.asd.lol.xd"), "valid")
+	assert.True(t, validHostname("randomtext"), "valid")
+
+	assert.False(t, validHostname("www.asd.lol.x-d"), "not valid")
+	assert.False(t, validHostname("192.168.1.1"), "not valid")
+	assert.False(t, validHostname("..a"), "not valid")
+	assert.False(t, validHostname("a..a"), "not valid")
+	assert.False(t, validHostname("123213.213"), "not valid") // look into this again
```
nikkolasg: have you decided?

network/address_test.go
```diff
@@ -63,3 +80,22 @@ func TestAddress(t *testing.T) {
 		assert.Equal(t, str.Public, add.Public())
 	}
 }
+
+// just a temporary test case
+func TestDNSNames(t *testing.T) {
+	assert.True(t, validHostname("myhost.secondlabel.org"), "valid")
```
nikkolasg: It's not a big deal, but you don't need to specify a third argument for `assert` and `require`, looking at the output we already know if it's valid or not and where is the invalid line, so it does not bring any additional useful information.

network/address_test.go
marcozov: How do you define these static (correct) values? IP addresses associated to a hostname might change

nikkolasg: We don't care about the validity of the relation between host name <-> ip address in the real world, we just want a representative mapping in our test, that's it.
` google.com <-> 8.8.8.8 ` for example.

nikkolasg: We must not test the behavior of `net.LookupHost` (as it's already tested by Golang) but only the behavior of the `NetworkAddress` function.

network/address.go
marcozov: I did it so that the returned IP address is always the same and testing would be easier (LookupHost returns the address in a random order). Otherwise, test should be change in such a way that the test case takes all the IP addresses associated to the hostname, and we could just see whether the IP address returned from NetworkAddress is included in those IP addresses of the test.

ineiti: I would put the sort into the test, it's not really needed in this method.

nikkolasg: Yes, we should limit the number of "hacks" like that that are only needed for testing. I would remove that line, and decide on a good way to test that function. Your proposal sounds fine to me, but it can be even simplified using the `Interface` approach mentioned below (the `TestResolver` could return an array of length one, so there would be no need to search through the array).

network/address_test.go
ineiti: @nikkolasg - I like the idea of a `Resolver`, however, `LookupHost` is in a package, not in a structure. So would you add another argument to `validHostname` that represents the `Resolver` which would wrap the `LookupHost`-method?

In fact, you could simply pass a `func LookupHost(host string) (addrs []string, err error)` to `validHostname` and either supply `net.LookupHost` (without brackets) or `LookupHostStub` that will be defined in the test.

marcozov: @ineiti why should I pass the function as argument to validHostname? This function does not really care about how the hostname is resolved.
Maybe you meant NetworkAddress?

network/address_test.go
marcozov: Yes, for now I do not include IDNs (i.e. hyphens/numbers in TLDs are not admitted)

network/address_test.go
ineiti: Oops - of corse. This is what I meant ;)

nikkolasg: I was just thinking of something more hacky-ish, having a non exported global variable `var res Resolver = NewGolangNetResolver()`. When doing the test, you just set that variable to `Resolver = newTestResolver()`. 
Of course, as @ineiti suggested, we can do that with just a simple `type resolv func(host string) (addrs []string,error)`, I just find less beauty in it, but I'm totally fine with it too.

marcozov: The fact is that such change (the additional parameter, representing how the hostname is resolved) would be needed for all the functions that rely on NetworkAddress() (like Host() or Port()). Does this really make sense?

ineiti: I don't care for debugging-variables. You can even make them private, then as long as you have a unit-test, you can overwrite it.

Again, I would not define the interface globally, but the lookup-function. So

```
var lookupHost func(host string)(addrs []string, error) = net.LookupHost
```

then you can overwrite that for the tests.

nikkolasg: +1 for @ineiti 's suggestion. You don't need to change any method's signature. Just change `net.LookupHost(host)` in `NetworkAddress()` func, by `lookupHost(host)`, as defined by @ineiti just above. The other functions don't need to know about that.

network/address_test.go
nikkolasg: Then I guess you can remove the comment :)

nikkolasg:
Another possible problem that occured to me: 
Now, *each time* one call `NetworkAddress()`, this method potentially calls the resolver, so potentially make a network call. Is it good ? 
Initially this method just to separate the "tls://" or "tcp://" part from the actual real network address. But now it does something much more and I fear that this will bring its lot of problem. For example, when running behind a proxy, this won't show. 
It would have been good if there was an explicit "Name" address and "IP" address maybe, with the former having a `Resolve() IPAddress` method where it's explicit that a network call will be operated. 
I don't mean to do anything necessarily by now but just wanted to raise the issue in case any of you have thought about it or has any ideas....

nikkolasg:
See all usages in our repo of this function. I'm feeling a bit anxious now about whether to directly integrate the resolving part in this function .... :)
```
cothority/cosi/check/check.go:116:			serverStr += s.Address.NetworkAddress() + "_"
cothority/cisc/cisc.go:145:	str := fmt.Sprintf("cisc://%s/%x", cfg.Cothority.RandomServerIdentity().Address.NetworkAddress(),
onet/simulation.go:128:			_, port, _ := net.SplitHostPort(sc.Roster.List[i].Address.NetworkAddress())
onet/app/server.go:301:		ln, err := net.Listen("tcp", binding.NetworkAddress())
onet/app/server.go:313:	conn, err := net.Dial("tcp", ip.NetworkAddress())
onet/app/server.go:316:		"can connect to", ip.NetworkAddress())
onet/app/server.go:321:	_, port, err := net.SplitHostPort(ip.NetworkAddress())
onet/local.go:303:	addr := network.NewTCPAddress(id.Address.NetworkAddress())
onet/network/address.go:117:	na := a.NetworkAddress()
onet/network/address.go:121:	h, _, e := net.SplitHostPort(a.NetworkAddress())
onet/network/address.go:136:	na := a.NetworkAddress()
onet/network/address.go:156:		"(^\\[::\\])", a.NetworkAddress())
onet/network/tcp.go:53:	netAddr := addr.NetworkAddress()
onet/network/tcp.go:263:	global, _ := GlobalBind(addr.NetworkAddress())
```
What do you think ?

Another idea: `NewDNSAddress` which takes the address and directly resolves it and cache the result, and we mark it as explicit that it makes a DNS query. 

Anyway, **again**, I don't necessarily want to do something **now** , just wanted to .. "open the debate" :)

ineiti:
Yes, that bothered me, too. My idea was to put this into `NewAddress` and then only store the IP. But this wouldn't catch when somebody doesn't use `NewAddress`. I'm OK for DNS-lookups in either `NewDNSAddress` or `NewAddress`.

marcozov:
@nikkolasg I see the point, but until now DNS was not even handled. With usual IP addresses ( 4 octet notation ) the behavior is not gonna change, so there should be no issue even if the lookup is in the code. In fact, the lookup is used only if the provided host does not have a proper IP address syntax ( 4 octet notation ).
Maybe it is more a matter of deciding whether we want/need some functions to handle **any** hostname (i.e. even if it is not resolved, but it is still well formed)? Because with this code, only hostnames that can be resolved can be used. But this might make sense if functions work directly with the IP addresses (which should be the case, since DNS was not handled). In this case I think that we could simply have NetworkAddress acting as before, and a new one that performs the DNS resolution

nikkolasg:
>  In this case I think that we could simply have NetworkAddress acting as before, and a new one that performs the DNS resolution

Mhh I like this in fact. It's true that `NetworkAddress` could also return the DNS Name as well, and if the caller wants an IP address absolutely, then he can call `Resolve()`. Stg like:
```go
func (a Address) Valid() bool {} // returns true as long as it's a valid DNS name or IP addr.
func (a Address) DNSOrIP() bool {} // returns true if it's a domain name or false
func (a Address) NetworkAddress() string {} // returns the address without the prefix "tcp://"
func (a Address) Resolve() string {} // returns the IP address, possibly add argument here for the resolver.
```

What do you think ? It avoids the issue of resolving always, and still provides a nice API to work with both types transparently.

marcozov:
@nikkolasg it seems ok. I will finish this tonight.

ineiti:
`s/DNSOrIP/IsHostname/`

network/address.go
```diff
@@ -70,6 +82,86 @@ func (a Address) NetworkAddress() string {
 	return vals[1]
 }
 
+// NetworkAddressResolved returns the network address of the address, but resolved.
+// That is: the hostname resolved and the port joined by a colon.
+// It returns an empty string if the address is not valid.
+func (a Address) NetworkAddressResolved() string {
```
nikkolasg: I'm not sure I understand the need for a distinct `NetworkAddressResolved` and `Resolve`: both are doing the resolution, one is only giving the ip address while the other gives the IP address and the port... Do we really need two distinct functions ? Which one are we going to use later ?  I find it confusing. 

Another question is whether `Resolve` should return another `Address` instead of a plain string ? Since `NetworkAddress` already has the purpose of extracting the ip address and `Port` the port, therefore kind of "removing" the need for the two distinct methods mentionned above.

PR #180, opened 2017-05-09
V1.0 merge

ineiti:
Merging things from master back into v1.0

One question: are the changes to the IDs too much of an API-change? There is one incompatible change: `Tree.Equals` got changed to `Tree.Equal`.

- change it back to `Tree.Equals`?
- prepare it for v1.1/v2.0?


ineiti:
Re-added `TreeID.Equals` (with an `s`) but using `log.Warn` for deprecated function.

nikkolasg:
> Re-added TreeID.Equals (with an s) but using log.Warn for deprecated function.

+1000 !!!

PR #182, opened 2017-05-11
Realpath 181

### LINKED ISSUE #181: Check for realpath
`libtest.sh` uses `realpath` that is not always installed. Look for it and use an alternative function that _might_ work.

ineiti:
Closes #181

### DIFF
--- a/app/libtest.sh
+++ b/app/libtest.sh
@@ -276,7 +276,7 @@ if ! which pcregrep > /dev/null; then
 	echo "*** WARNING ***"
 	echo "Most probably you're missing pcregrep which might be used here..."
 	echo "On mac you can install it with"
-	echo "brew install pcre"
+	echo -e "\n  brew install pcre\n"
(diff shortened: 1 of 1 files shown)


app/libtest.sh
```diff
@@ -276,11 +276,23 @@ if ! which pcregrep > /dev/null; then
...
 fi
 
+if ! which realpath > /dev/null; then
+	echo "*** WARNING ***"
+	echo "Most probably you're missing realpath which might be used here..."
+	echo "On mac you can install it with"
+	echo -e "\n  brew install coreutils\n"
+	echo "Not aborting because it might work anyway."
+	echo
+	realpath() {
```
nikkolasg: Out of curiosity, why do we need realpath ?

ineiti: it removes links and `..`: http://man7.org/linux/man-pages/man3/realpath.3.html

PR #183, opened 2017-05-22
[V2] Using kyber.v1

nikkolasg:
This works bring kyber.v1 to Onet.
I have to say that I'm still quite opposed to this approach. The two main problems here:
+ Naming issue. We had a `network.Suite` previously. I've made that an interface defining what the `onet` suite's capabilities (`Group`, `hash.Hash` and `Cipher`). 
Now, the concrete implementation (edwards25519) has to be named stg else ! I've settled for `network.S` as it's short and concise. I still find that more confusing than anything else to have `network.Suite` and `network.S` (regardless of the name).
+ Capabilities issue: (`Group`, `hash.Hash` and `Cipher`). These are the only one I know we are using but maybe others are needed. I'm not sure and I **can't** be sure since onet is the basis of many ongoing projects. So `network/` defines a `Suite` with much more than intented to. That was exactly the thing we wanted to solve with the whole `Suite` restructuring. It does not make sense for me...


network/encoding.go
```diff
@@ -4,24 +4,31 @@ import (
...
 
 /// Encoding part ///
 
-// Suite used globally by this network library.
+// Suite functionalities used globally by the network library.
+type Suite interface {
+	kyber.Group
+	Hash() hash.Hash
+	Cipher(key []byte, options ...interface{}) kyber.Cipher
+}
```
nikkolasg: Here is the relevant part

ineiti:
The branch/merge-to should be changed to:
branch: v2 -> use_kyber
merge-to: master -> pre_v2

Discussing yesterday with @bford we don't want to have a `v2` before it's ready. So the branch where we do development is `pre_v2`. And this change should be in branch `use_kyber` that PRs against `pre_v2`.

nikkolasg:
I would have avoided using directly the v2 branch too but the same problem appeared in the crypto v1 branch: I can't use `gopkg.in` if the `vXX` branch does not exists. That means I could not have tested onet this way, and same for cothority which should depend on onet.v2 so, I'm not sure how you want to resolve this... This PR is mainly to try out how it looks in onet and cothority, not necessarily merging now.

ineiti:
True - and gopkg doesn't have a `pre`-functionality. Post-v2 will be something different than `gopkg`...

So `v2` it is then for `onet`, `v1` for `kyber`.

Put at least a big warning in the `README.md`-file to indicate it's not yet released.

ineiti:
With regard to the definition of `network.Suite` - what does `onet` use? 
- `Suite` - for sure
- `Hash` - do we hash in `onet`?
- `Cipher` - where do we use a cipher?

IIRC, the only point why we have to define a `Suite` in `network` is for marshalling/unmarshalling. So we could say that only a `Group` is needed in `onet/network`.

nikkolasg:
Ok I've removed the `Cipher` method and `onet` compiles. But the `Hash` method is required by `key.NewKeyPair` in kyber, which is used in `onet` (`local.go` and `simulation.go`) .    :/

ineiti:
Would it make sense to remove `Hash` from `NewKeyPair`?

jeffallen:
Closing this PR for now because we are renaming this branch to kyber.

PR #191, opened 2017-05-31
added registermessages

### LINKED ISSUE #190: network.RegisterMessages
There are quite some places where I do stuff like:
```
	for _, m := range []interface{}{
		// - Data structures
		&SkipBlockBunch{},
		&SBBStorage{},
	} {
		network.RegisterMessage(m)
	}
```

But I'd love to do simply
```
        network.RegisterMessages(&SkipBlockBunch{}, &SBBStorage{})
```

ineiti:
Closes #190

### DIFF
--- a/network/encoding.go
+++ b/network/encoding.go
@@ -78,6 +78,16 @@ func RegisterMessage(msg Message) MessageTypeID {
 	return msgType
 }
 
+// RegisterMessages is a convenience function to register multiple messages
+// together.
+func RegisterMessages(msg ...Message) []MessageTypeID {
+	var ret []MessageTypeID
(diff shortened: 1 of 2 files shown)


network/encoding.go
```diff
@@ -78,6 +78,16 @@ func RegisterMessage(msg Message) MessageTypeID {
 	return msgType
 }
 
+// RegisterMessages is a convenience function to register multiple messages
+// together.
```
nikkolasg: We could add in the comment 1) what is the returned value, and 2) specifying what happens in case multiple `Message` are the same. Namely nothing, the same ids will be returned multiple times.

PR #193, opened 2017-06-13
changed log-lvl to 3

ineiti:
lvl1 - test-messages and important messages from apps
lvl2 - important messages from services
lvl3 - messages that might help debug
lvl4 - debugging only in special cases
lvl5 - spamming debug-lvl and huge messages

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -199,7 +199,7 @@ func newMessageProxyStore(disp network.Dispatcher, proc network.Processor) *mess
 		io := newIO()
 		pstore.protos = append(pstore.protos, io)
 		disp.RegisterProcessor(proc, io.PacketType())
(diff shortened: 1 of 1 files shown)


nikkolasg:
Oh damn... Another race. 
I think we should build something from scratch with clearly compartimentalized modules if we do the `Graph` overlay. We would avoid having tons of mutex everywhere in the same file like we do for `Overlay`, it's not the first time we have surprises with this one.
I'm just waiting for your decision on that (merge or not ?) For me, it's LGTM

ineiti:
Nah, it's a real bug: https://github.com/dedis/onet/pull/195 - can you merge the 195, then I re-launch travis here.

PR #194, opened 2017-06-13
corrected bug

### LINKED ISSUE #169: Bug in onet/websocket
Now: if `onet.Client.Send` is called on a `NewClientKeep`, the connection will be kept even if we have a different command. So:

```
c := NewClientKeep("Skipchain")
c.Send(dst, "StoreSkipChain", sc)
c.Send(dst, "GetUpdateChain", last)
```

does not work, because the first send makes a connection, and the second send uses the same connection, connected to the same endpoint, which is still `StoreSkipChain` in that case.

Wanted: also store endpoint used (`path`) and re-connect if the path is different.

christophebertrand:
closes #169

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -227,7 +227,7 @@ func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]b
 	}
 	defer func() {
 		if !c.keep {
-			if err := c.Close(dest); err != nil {
+			if err := c.Close(); err != nil {
 				log.Errorf("error while closing the connection to %v : %v\n", dest, err)
 			}
 		}
@@ -289,17 +289,15 @@ func (c *Client) SendToAll(dst *Roster, path string, buf []byte) ([][]byte, Clie
 	return msgs, NewClientError(err)
 }
 
-// CloseAll sends a close-command to all connections and returns nil if no errors
-// occurred or all errors encountered concatenated together as a string.
-func (c *Client) CloseAll() error {
+// Close sends a close-command to all open connections and returns nil if no
+// errors occurred or all errors encountered concatenated together as a string.
+func (c *Client) Close() error {
 	var errstrs []string
-	for _, conn := range c.connections {
-		conn.WriteMessage(websocket.CloseMessage, nil)
-		if err := conn.Close(); err != nil {
+	for dest := range c.connections {
+		if err := c.closeConn(dest); err != nil {
 			errstrs = append(errstrs, err.Error())
 		}
 	}
-	c.connections = make(map[destination]*websocket.Conn)
 	var err error
 	if len(errstrs) > 0 {
 		err = errors.New(strings.Join(errstrs, "\n"))
@@ -307,8 +305,8 @@ func (c *Client) CloseAll() error {
 	return err
 }
 
-// Close sends a close-command to the connection.
-func (c *Client) Close(dst destination) error {
+// closeConn sends a close-command to the connection.
+func (c *Client) closeConn(dst destination) error {
 	conn, ok := c.connections[dst]
 	if ok {
 		delete(c.connections, dst)
--- a/websocket.go
+++ b/websocket.go
@@ -227,7 +227,9 @@ func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]b
 	}
 	defer func() {
 		if !c.keep {
-			c.Close(dest)
+			if err := c.Close(dest); err != nil {
+				log.Errorf("error while closing the connection to %v : %v\n", dest, err)
+			}
 		}
 	}()
 	if err := conn.WriteMessage(websocket.BinaryMessage, buf); err != nil {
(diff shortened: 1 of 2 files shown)


websocket.go
```diff
@@ -220,18 +223,18 @@ func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]b
 		if err != nil {
 			return nil, NewClientError(err)
 		}
-		c.si = dst
+		c.connections[dest] = conn
 	}
 	defer func() {
 		if !c.keep {
-			c.Close()
+			c.Close(dest)
```
ineiti: This should also remove `conn` from the `c.connections`.

nikkolasg:
@christophebertrand it's not related to your work but... as written in the issue, I have to express my feelings about this approach: opening a new connection for basically each endpoint looks bad.  The day we add TLS, we're gonna have to wait 1 sec for each TLS handshake for each message we want to send... 

Nevertheless.... this PR needs also to tackle the management of the connections, I believe the websocket library may have a readtimeout property or stg similar (Ping/Pong mechanism?) to look for.

ineiti:
For the moment we're stuck with this solution as we want to stay compatible within `v1`. Feel free to open an issue for `v2`.

websocket.go
christophebertrand: Isn't this done [here](https://github.com/dedis/onet/pull/194/files/791fa463220631c03ba26654e1b8bb1f250bdf3d#diff-8990d2c9c7f1690e60641d425c8932f7R294)?

nikkolasg:
Please see my comment on the issue. There is a way to keep the API the same and still re-use the same connection. https://github.com/dedis/onet/issues/169#issuecomment-308402607

websocket.go
ineiti: True - now I wonder whether `c.Close()` should not close _all_ connections... This would make more sense to me. What do you think?

ineiti:
> Please see my comment on the issue. There is a way to keep the API the same and still re-use the same connection. #169 (comment)

There you propose two changes in the API:
* you propose to change the path
* the message has to be sent using `network.Marshal` instead of `protobuf.Marshal`

This, of course, will break existing code, which is not OK before v2, IMHO.

IIRC, we chose to have full paths so we don't have to use `network.Marshal`, but so we can use unmodified protobuf-messages.

websocket.go
christophebertrand: If we close all connections how should errors be handled? Just return the last error returned by the connection or `nil` if no error occurred. 
Also now we don't handle the error returned by `c.Close()`, shouldn't the client do something if an error is returned?

nikkolasg: I think there should be a `Close()` closing all connections in addition to `Close(dest)`. Users will most likely call `Close()` in any cases.
Indeed, `c.Close()` error is not handled: it probably should be ;)

ineiti: Proposition:
- here: only close one connection, delete it from the map
- `Client.Close` - close all connections, return the `string`-concatenated error-messages, if any

christophebertrand: What would be the best way to handle the error?

ineiti: You can:
1. return a slice of errors (which might be empty)
2. return a concatenated string of errors
3. return the first error and then either stop on error or go on anyway
4. only `log.Error` the errors
5. Ignore the errors altogether

Who cares if a connection-*close* has an error - the *opening* error is important, as is the *writing* error. But not the closing error. So I propose to:
- `log.Error` an eventual error in the *closing* part in `Client.Send` - this will log the error on stderr, but make the `Client.Send` method succeed *if* the open and write succeeded, of course.
- return a concatenated string of errors (or nil if no error) in `Client.Close`

websocket.go
```diff
@@ -287,13 +289,29 @@ func (c *Client) SendToAll(dst *Roster, path string, buf []byte) ([][]byte, Clie
 	return msgs, NewClientError(err)
 }
 
+// CloseAll sends a close-command to all connections.
```
ineiti: Add in the comment what the returned error contains.

ineiti:
I think there was some misunderstanding. `Client.Close(conn)` doesn't make any sense - `Close` is public, but nobody else can give it a valid argument. So for minimal code-duplication:

- rename `Close` to `closeConn`
- rename `CloseAll` to `Close` and use `closeConn` instead of closing the connections on its own

websocket.go
```diff
@@ -284,13 +289,29 @@ func (c *Client) SendToAll(dst *Roster, path string, buf []byte) ([][]byte, Clie
...
-// Close sends a close-command to the connection.
+// Close sends a close-command to all open connections and returns nil if no
+// errors occurred or all errors encountered concatenated together as a string.
 func (c *Client) Close() error {
-	if c.si != nil && c.conn != nil {
-		c.si = nil
-		c.conn.WriteMessage(websocket.CloseMessage,
-			nil)
-		return c.conn.Close()
+	var errstrs []string
```
ineiti: Needs to take the `Client`-lock.

websocket.go
```diff
@@ -220,18 +223,20 @@ func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]b
 		if err != nil {
 			return nil, NewClientError(err)
 		}
-		c.si = dst
+		c.connections[dest] = conn
 	}
 	defer func() {
 		if !c.keep {
-			c.Close()
+			if err := c.Close(); err != nil {
```
ineiti: use `Client.connClose` else it will deadlock.

ineiti:
Wow - that was a long one. Thanks for your stamina! I'm OK to merge - @nikkolasg ?

PR #202, opened 2017-07-06
Add configuration files and command line 

nikkolasg:
This PR brings the functionality of having configuration files to the onet framework. Configuration files can be TOML. JSON or anything that can be read be the [viper framework](https://github.com/spf13/viper).
Moreover, it adds a wrapper around different sources of configuration so that the command line interface can override any configuration file item also. It implements a wrapper around the `urfave/cli` interface. 
@ineiti maybe later, we could have the skipchain update the configuration in this wrapper too ;)


bford:
Good to learn about this Viper framework; sounds promising.

But one question: does it have any support for dealing with cryptographic secrets (e.g., locally-stored private keys) in the most secure OS-specific fashion, e.g., using Keychain on Mac OS and the equivalent on other OSes?  Or are there any other "configuration frameworks" or other Go libraries available that provide some kind of cross-platform support for OS-specific key-management frameworks?

If not, this might be something we really need to add to Onet or Kyber at some point...  (not necessarily right now but we should before too long; it's definitely not consistent with the state-of-the-art security practice just to stash private keys in plain files in home directories anymore...)

nikkolasg:
No viper does not have that functionality but that's not the purpose of that PR. Here the sole purpose is to have a generic configuration item to be able to retrieve *non-sensible* configuration items from a file such as "server.ip = 0.0.0.0" or "identity.skipchain.blockmax = 10m". Such a config file could easily be added to `/etc/cothorityd/` later.

@ineiti already asked me the same question about the secret keys and I'm planning to attack that subject just after this one ;) I'll try to see what is available in Go for OS secure storage. I think we'll have to be flexible and allow a mix of different kind of storages (for simulations I envision it'll be difficult to use the OS-secure storage always).

config/source.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return false
+}
+
+// TypedSource adds additional functionality to a Source. Namely, it provides
+// wrapper function to cast any value to a certain type such as int,
+// time.Duration, etc. If the key is not found, or an error happened during the
+// conversion, then the default value of the type is returned.
+type TypedSource struct {
+	Source
```
nikkolasg: Do anybody know if and how it is possible to do a type over an `interface` ? 
I tried and I got compilation error, and tried this approach after 10mn of trying around. 
`type TypedSource Source` works but when you want to define methods on `TypedSource`, go does not like to have an interface type in the receiver...

bford:
Sounds good, thanks Nicolas!

config/cli_source.go
```diff
@@ -0,0 +1,143 @@
...
+// Finally, CliSource implements a feature to be able to define any key/value
+// pair on the command line. To use this feature, you need to register a
+// GenericFlag to the `cli.App` or `cli.Command` such as
+//
+//    Flags: []cli.Flag{Name: "generic", Value: &config.GenericFlag{}}
+//
+// This way, one can write any key/pair on the command line using the following
+// format:
+//
+//   ./MyApp --generic "globalKey=value" <command> --generic "key=value"
```
ineiti: I got used to this 'sub-command flags' handling that urfave/cli does. I'm not sure we need the second `--generic`, too.

config/cli_source.go
```diff
@@ -0,0 +1,143 @@
...
+		return ErrGenericFlagFormat
+	}
+	p := &pair{strs[0], strs[1]}
+	g.pairs = append(g.pairs, p)
+	return nil
+}
+
+// String implements the generic flag interface from urfave/cli
+func (g *genericFlag) String() string {
+	return "generic flag setting: --generic \"<key>=<value>\""
```
ineiti: This should use `GenericFlagName`.

config/viper_source.go
```diff
@@ -0,0 +1,36 @@
...
+//    viper.SetConfig("name")
+//    viper.SetConfigPath(".")
+//
+func NewViperSource() *ViperSource {
+	return &ViperSource{viper.GetViper()}
+}
+
+// Defined returns true if the key is defined in the configuration file
+func (v *ViperSource) Defined(key string) bool {
+	return v.v.IsSet(key)
```
ineiti: We could go with viper's definition and do a `s/Defined/IsSet/`.

config/viper_source_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestViperSource(t *testing.T) {
+	viper.SetConfigType("TOML")
+	cosiRate := "10s"
+
+	var tomlExample = []byte(`
+[server]
+ip = "0.0.0.0"
+
+[cosi]
+rate = "10s"
```
ineiti: We need to find a better example - the rate is given in `s**-1`, not `s`: is this 10 signatures per second or one signature every 10 seconds?

config/viper_source.go
nikkolasg: It's a taste issue at this point whether we prefer `IsSet` or `Defined` because I'll anyway need to define `ViperSource` since I don't want the `GetString` but just the `String` method.

config/cli_source.go
nikkolasg: It depends if the "creator" of the application sets a generic flag in the sub-command or not. I just wanted to put it there so the doc states that it's possible to put it afterwards too. Whether you want or not is up to you ;)

nikkolasg:
What would you have done for `MultiSource` ?
I did so because I prefer having each functionality clearly separated, and be able to compose / change them at will afterwards.

ineiti:
For multi-platform keyring, I found https://github.com/tmc/keyring - but it lacks Windows support. Only MacOSX/gnome.

ineiti:
I can't think of any other source of configurations, so I suppose we'll only ever have `viper`/`urfave`-implementations. And the order seems to be quite standard: cli beats config-file beats default.

Anyway - nice structure for the configuration!

ineiti:
@nikkolasg - when thinking of private/public keys - I think we need to differentiate two sets of private keys:
1. the ones coming from the apps (`cisc`, `pop`, `pulsar`, `scmgr`)
2. the ones being used by the services

For 1. it makes sense to store them in a keyring (I would say this makes a nice `[conode]`-configuration option), while 2. will be run on servers and in multiple instances anyway, so no need to store those in the keychain.

What do you think?

nikkolasg:
I've created an issue specifically for that, please comment in https://github.com/dedis/onet/issues/203

nikkolasg:
>  even though it might be over-engineered a bit using the MultiSource-approach

I merged both `MultiSource` and `TypedSource` into one `HubSource` so complexity is reduced.

jeffallen:
Closing because we want to do this work on the branch called "kyber".

jeffallen:
This PR had to be closed to enable us to delete v2 branch. However, this most certainly belongs in onet's kyber branch. I will need to create a new PR to do that.

PR #213, opened 2017-08-22
Portscan 212

### LINKED ISSUE #212: portscan not working anymore
the previous `php`-call possible with `whatsmyip.org` does not work anymore. I propose we replace it with a simple service on `dedis.ch` that listens on a port. If a host sends a number to that port, the service will try to connect back to the host at the sent port-#. This avoids abuse of the system.

ineiti:
Closes #212

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -1,12 +1,10 @@
 package app
 
 import (
-	"bytes"
 	"errors"
 	"io/ioutil"
 	"net"
 	"net/http"
-	"net/url"
 	"os"
 	"os/user"
 	"path"
@@ -24,6 +22,8 @@ import (
 	// CoSi-protocol is not part of the cothority.
 
 	// For the moment, the server only serves CoSi requests
+	"fmt"
+
 	"gopkg.in/dedis/crypto.v0/abstract"
 	crypconf "gopkg.in/dedis/crypto.v0/config"
 )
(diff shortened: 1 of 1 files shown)


nikkolasg:
Should we not merge that first into master, then v1.0 ? Since that is beneficial for all versions.

ineiti:
yes, but I'm creating docker-images:

https://hub.docker.com/r/dedis/conode/

and this depends on v1 - so it was simpler to make this on v1 - but I'll do the backport. Or forwardport?

app/server.go
```diff
@@ -318,41 +324,33 @@ func tryConnect(ip, binding network.Address) error {
...
+	buff, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
 	if err != nil {
 		return err
 	}
 
-	if !bytes.Contains(buffer, []byte("1")) {
-		return errors.New("Address unreachable")
+	res := string(buff)
+	if strings.Compare("Open", res) != 0 {
```
nikkolasg: It's nitpicking but it seems overly complicated for a string comparison... why not simply `res == "Open"` ?

nikkolasg: Even golang says it's clearer and faster to use the operator == in [the doc](https://golang.org/pkg/strings/#Compare).

ineiti: oops - I had java on my mind...

PR #223, opened 2017-09-01
fail if receiving-channel is full

### LINKED ISSUE #219: onet-channels block if there are too many messages waiting
In line

https://github.com/dedis/onet/blob/master/treenode.go#L168

onet blocks if the message-queue gets too big. We can either increase the queue-size in `RegisterChannel` or put the above call to `Send` in a go-routine.

ineiti:
Closes #219

### DIFF
--- a/treenode.go
+++ b/treenode.go
@@ -56,11 +56,14 @@ type TreeNodeInstance struct {
 	configMut sync.Mutex
 }
 
-// aggregateMessages (if set) tells to aggregate messages from all children
-// before sending to the (parent) Node
-// https://golang.org/ref/spec#Iota
 const (
-	AggregateMessages = 1 << iota
+	// AggregateMessages (if set) tells to aggregate messages from all children
+	// before sending to the (parent) Node
+	AggregateMessages = 1
+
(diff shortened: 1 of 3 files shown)


treenode.go
```diff
@@ -368,10 +378,16 @@ func (n *TreeNodeInstance) DispatchChannel(msgSlice []*ProtocolMsg) error {
...
 		for _, msg := range msgSlice {
-			out := n.channels[mt]
+			out := reflect.ValueOf(n.channels[mt])
 			m := n.reflectCreate(to.Elem(), msg)
 			log.Lvl4(n.Name(), "Dispatching msg type", mt, " to", to, " :", m.Field(1).Interface())
-			reflect.ValueOf(out).Send(m)
+			if out.Len() < out.Cap() {
+				out.Send(m)
+			} else {
+				return fmt.Errorf("channel too small for msg %s in %s: "+
```
nikkolasg: I would even panic in this case. What do you think ?

ineiti: that would mean that a rogue node could panic other nodes... I prefer not to panic there. Ideally it should stop the protocol, but I did not implement that.

PR #224, opened 2017-09-12
Add input validation

### LINKED ISSUE #189: Input validation
Under certain circumstances, a conode will fail if it receives illegal input in either the binary or the websocket-port. Unfortunately I don't have a specific example, but we should look at the input-paths from both the binary and the websocket-receivers and make sure that errors are caught, such as:
- illegal messages
- too big messages

So all possible and imaginable error-handling should be done. An easy random-message-catcher would be to add a sha256 to all messages sent over the network.

Another way would be to push (finally) the TLS-implementation for the binary receiver.

And finally we could do some fuzzy-testing.

ineiti:
* Add check of the packet size

Closes #189

### DIFF
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -18,6 +18,8 @@ import (
 // sent.
 var readTimeout = 1 * time.Minute
 
+const maxPacketSize = 10 * 1024 * 1024
+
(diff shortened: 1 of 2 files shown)


nikkolasg:
Nice !

PR #225, opened 2017-09-14
Adding stable script

ineiti:
Last of kyber and cothority to get a stable-script. Changes include:

- adding _deprecated_ warnings for `crypto.*Pub*`
- clean up `simul/test_simul`
- merged channel-restriction in `TreeNodeInstance` from `v1` back to `master`
- for `v1` there will be a lot of patches that didn't make it from `master` to `v1`...
- bumping version from `v1.0` to `v1.2`
- fixing tests so they pass Travis that doesn't support IPv6


.travis.yml
```diff
@@ -5,18 +5,14 @@ go:
...
 script:
   - make test
 
-after_success:
-  - $GOPATH/bin/goveralls -coverprofile=profile.cov -service=travis-ci -repotoken $COVERALLS_TOKEN
-
 notifications:
   email: false
+
+sudo: required
```
nikkolasg: Why requiring sudo ? Apparently, on this [line](https://github.com/dedis/onet/pull/225/files#diff-891d77d087977450358304892d384a08L11) you already check if sudo is there or not and act accordingly.

ineiti: True - there are other tests that require `sudo` - should I add the same `t.Skip` there?

nikkolasg: I don't know. Why do we need sudo at all ? when I run `go test` it uses sudo on my machine ?

ineiti: Because of

https://github.com/dedis/onet/blob/4b2d0e5f1133fc8c1639303c5f025bb55afb5400/simul/platform/localhost.go#L166

but perhaps `sudo` is not really needed there. In fact, you tell me: this was for your pairing-based crypto libraries - can you install them also as the local user?

nikkolasg: I would prefer to assume the script knows better if it needs sudo or not and if it can run with or without sudo instead of directly forcing him as sudo. 
So I think we can remove the sudo necessary condition here.

PR #227, opened 2017-09-15
exporting MaxPacketSize

ineiti:
If you need bigger packets, you should be able to increase the size.

Rant: if we would have put a hash or some magic bytes we wouldn't need this ;)

### DIFF
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -18,7 +18,10 @@ import (
 // sent.
 var readTimeout = 1 * time.Minute
 
-const maxPacketSize = 10 * 1024 * 1024
+// MaxPacketSize limits the amount of memory that is allocated before a packet
+// is checked and thrown away if it's not legit. If you need more than 10MB
+// packets, increase this value.
+var MaxPacketSize = Size(10 * 1024 * 1024)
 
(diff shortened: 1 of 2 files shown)


nikkolasg:
Rant: you should not need to send more than 10mb at once ;)
https://developers.google.com/protocol-buffers/docs/techniques
https://developers.google.com/protocol-buffers/docs/techniques

ineiti:
Tell that the guys from LCA1... and they're our customers, so they're right ;)

nikkolasg:
Well what can I say. I think it's again one more thing that will shoot us in the foot/head/whatever when doing simulations for example. Also if some have different values, it may causes weird interactions: "why my packet is not received on your conode?" "Oh I can only send up to 10Mb on yours, and on mine 100Mb" " let's just switch to 100Mb then and we'll both be happy" ... 
The solutions I see are to either a) doing the split up ourself and recontruct ourself at the other end, b) indicate that you can not send more than 10mn through onet in one packet. 
I tend for b) as it leads to more safe strategy thinking in the long term in my opinion.
In the meantime, feel free to merge...

PR #229, opened 2017-09-27
changing the way the hash is calculated

ineiti:
The onet-schnorr-signature is a not-eddsa-compatible schnorr signature. For v2 we'll have to remove it. For previous versions, I propose we change the hash to the below function, so that the javascript-implementation from the epfl e-voting project can easily interact with the cothority.

### DIFF
--- a/crypto/schnorr.go
+++ b/crypto/schnorr.go
@@ -1,76 +1,45 @@
 package crypto
 
 import (
-	"errors"
-
 	"gopkg.in/dedis/crypto.v0/abstract"
-	"gopkg.in/dedis/crypto.v0/random"
+	"gopkg.in/dedis/crypto.v0/sign"
 )
 
 // SchnorrSig is a signature created using the Schnorr Signature scheme.
 type SchnorrSig struct {
-	Challenge abstract.Scalar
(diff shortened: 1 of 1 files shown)


nikkolasg:
Euhh Why is still there  ? I thought this was removed a loooong time ago in favor of the one in the crypto library ? The best move would be to remove that once and for all...

ineiti:
And then use https://github.com/dedis/kyber/blob/v0/sign/schnorr.go ?

nikkolasg:
Absolutely. Does that seem possible ?

ineiti:
How about that? So the existing code continues to be usable, but it calls the appropriate crypto-libraries.

nikkolasg:
Great ! Do you think it's worth putting a notice "deprecated- must use crypto library schnorr directly" or stg?
If not, feel free to merge.

ineiti:
In fact nobody is using that `SchnorrSig.Challenge` anyway...

ineiti:
And dang - it breaks a test:

https://github.com/dedis/cothority/blob/master/identity/api_test.go#L369

But that's probably worth it, what do you think?

PR #231, opened 2017-11-03
Libtest search

jeffallen:


app/libtest.sh
```diff
@@ -197,14 +197,22 @@ buildDir(){
...
 buildConode(){
-	local incl="$@"
-    local pkg=$( realpath $BUILDDIR | sed -e "s:$GOPATH/src/::" )
+    local incl="$@"
+    if [ -z "$incl" ]; then
+	echo "buildConode: No import paths provided. Searching."
+	for i in service ../service
+        do
+	    if [ -d $APPDIR/$i ]; then
+		local pkg=$( realpath $APPDIR/$i | sed -e "s:$GOPATH/src/::" )
```
ineiti: should we use bash-replacements here? Sthg. like `${$GOPATH/src/#}`? I'm not sure about the `#` part...

PR #232, opened 2017-11-03
Documentation improvements.

jeffallen:


tree.go
```diff
@@ -60,18 +60,18 @@ func (tId TreeID) IsNil() bool {
 	return tId.Equal(TreeID(uuid.Nil))
 }
 
-// NewTree creates a new tree using the entityList and the root-node. It
+// NewTree creates a new tree using the given entityList and root. It
 // also generates the id.
-func NewTree(el *Roster, r *TreeNode) *Tree {
-	url := network.NamespaceURL + "tree/" + el.ID.String() + r.ID.String()
+func NewTree(entityList *Roster, root *TreeNode) *Tree {
```
ineiti: that's an old name - you can directly replace it with `r` or `roster`.

jeffallen: done

PR #236, opened 2017-11-16
testing that a new protocol is correctly tied to a service

### LINKED ISSUE #235: CreateProtocol needs to respect ServiceID
Now: if a protocol that has been started by a service creates a new protocol, this new protocol will not be attached to the service and so it will not call the `NewProtocol` method of the service, making it impossible for the service to set configuration options of this protocol.

Wanted: The `CreateProtocol` method of `TreeNodeInstance` should copy the `ServiceID` so that the new protocol is also attached to the service.

ineiti:
Closes #235

### DIFF
--- a/treenode.go
+++ b/treenode.go
@@ -673,7 +673,7 @@ func (n *TreeNodeInstance) SendToChildrenInParallel(msg interface{}) error {
 // returns it with any error that might have happened during the creation. This
 // protocol is only handled by onet, no service are "attached" to it.
 func (n *TreeNodeInstance) CreateProtocol(name string, t *Tree) (ProtocolInstance, error) {
-	pi, err := n.overlay.CreateProtocol(name, t, NilServiceID)
+	pi, err := n.overlay.CreateProtocol(name, t, n.Token().ServiceID)
 	return pi, err
 }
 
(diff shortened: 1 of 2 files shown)


treenode.go
```diff
@@ -673,7 +673,7 @@ func (n *TreeNodeInstance) SendToChildrenInParallel(msg interface{}) error {
 // returns it with any error that might have happened during the creation. This
 // protocol is only handled by onet, no service are "attached" to it.
 func (n *TreeNodeInstance) CreateProtocol(name string, t *Tree) (ProtocolInstance, error) {
-	pi, err := n.overlay.CreateProtocol(name, t, NilServiceID)
+	pi, err := n.overlay.CreateProtocol(name, t, n.Token().ServiceID)
```
nikkolasg: Lol. That was easy ;)

nikkolasg:
LGTM

treenode.go
```diff
@@ -673,7 +673,7 @@ func (n *TreeNodeInstance) SendToChildrenInParallel(msg interface{}) error {
 // returns it with any error that might have happened during the creation. This
```
jeffallen: Please update comment to new behavior, also s/instantiates/instantiate/

treenode.go
ineiti: First I wanted to submit only this line - then I thought that I have 2 engineers that will tell me to write some test...

PR #241, opened 2017-11-24
Adding failing test case for Roster.GenerateNaryTreeWithRoot

ChristianMct:
There is an unexpected behavior in  `Roster.GenerateNaryTreeWithRoot` method.

It is the only one in the `Roster` to `Tree` function family to create a new `Roster` (with a new `RosterID`), which seems undesirable in most cases.


ineiti:
Closing, as I opened issue #242

PR #250, opened 2017-11-29
added OnCurrent method and tests

### LINKED ISSUE #248: Regression in GenerateNaryTreeWithRoot
After fixing `GenerateNaryTreeWithRoot` some protocols don't work anymore, because they presume that the root-node is at position `Tree.List[0]`.

Fix:
- Add `GenerateNaryTreeWithRootOnCurrent`, because it is a method of a `Roster`, and the `OnCurrent` method will work on the current roster
- For v2018, change the behaviour of `GenerateNaryTreeWithRoot` and remove the `OnCurrent` method

### LINKED ISSUE #249: Roster extension
Add a method:
`Roster.NewWithRoot`
that creates a new roster with the given root at position '0'. This makes it possible to simulate the old behaviour of `CreateNaryTreeWithRoot` in `v1`.

ineiti:
Closes #248

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -504,10 +504,31 @@ func (ro *Roster) GenerateBigNaryTree(N, nodes int) *Tree {
 // GenerateNaryTreeWithRoot creates a tree where each node has N children.
 // The root is given as an ServerIdentity. If root doesn't exist in the
 // roster, `nil` will be returned.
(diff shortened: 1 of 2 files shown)


ineiti:
Not needed in master - let it break!

jeffallen:
This really was needed on tip, because other users were depending on the original behavior.

PR #255, opened 2017-12-04
reviving old crypto/ key conversions

ineiti:
Some of the useful key conversions (hex- and base64 to string and back) got lost when deleting the `crypto/`-directory. Most probably they'll be used in the apps, so I copied them to the apps-directory.

### DIFF
--- a/app/key.go
+++ b/app/key.go
@@ -1,0 +1,212 @@
+package app
+
+import (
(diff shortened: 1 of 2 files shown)


jeffallen:
See kyber/util/encoding for where these went.

PR #261, opened 2017-12-08
Add boltdb to onet server

kc1212:
Prerequisite for https://github.com/dedis/cothority/pull/964

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,7 +1,6 @@
 package onet
 
 import (
-	"errors"
 	"os"
 	"os/user"
 	"path"
@@ -15,10 +14,11 @@ import (
 
 // Context represents the methods that are available to a service.
 type Context struct {
-	overlay   *Overlay
-	server    *Server
-	serviceID ServiceID
-	manager   *serviceManager
+	overlay    *Overlay
+	server     *Server
+	serviceID  ServiceID
+	manager    *serviceManager
+	bucketName string
 	network.Dispatcher
 }
 
@@ -30,6 +30,7 @@ func newContext(c *Server, o *Overlay, servID ServiceID, manager *serviceManager
 		server:     c,
 		serviceID:  servID,
 		manager:    manager,
+		bucketName: ServiceFactory.Name(servID),
 		Dispatcher: network.NewBlockingDispatcher(),
 	}
 }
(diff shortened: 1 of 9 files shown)


context.go
```diff
@@ -130,79 +129,61 @@ var testContextData = struct {
 }{service: make(map[string][]byte, 0)}
 
 // Save takes an identifier and an interface. The interface will be network.Marshaled
-// and saved under a filename based on the identifier. An eventual error will be returned.
-// If contextDataPath is non-empty, the destination is a file: it will be created
-// with rw-r----- permissions (0640). If the file already exists, it will be overwritten.
-//
-// The path to the file is chosen as follows:
```
jeffallen: Kind of too bad to lose this documentation of where the files are... did it move someplace else?

context.go
```diff
@@ -130,79 +129,61 @@ var testContextData = struct {
...
-		}
-	} else {
-		var err error
-		buf, err = ioutil.ReadFile(c.absFilename(id))
-		if err != nil {
-			return nil, err
-		}
+	bucketName := ServiceFactory.Name(c.ServiceID())
+	c.manager.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(bucketName))
```
jeffallen: chaining (i.e. v := tx.Bucket().Get()) would seem to be reasonable here.

context.go
```diff
@@ -130,79 +129,61 @@ var testContextData = struct {
...
-	fname := c.absFilename(id)
-	if getContextDataPath() == "" {
-		testContextData.Lock()
-		testContextData.service[fname] = buf
-		testContextData.Unlock()
-		return nil
-	}
-
-	return ioutil.WriteFile(fname, buf, 0640)
+	bucketName := ServiceFactory.Name(c.ServiceID())
```
ineiti: You could put `bucketName` in `Context`, then it's available for `Save` and `Load`.

context_test.go
```diff
@@ -20,60 +22,48 @@ type ContextData struct {
...
 	wg.Add(nbr)
 	for i := range c {
 		go func(i int) {
-			testLoadSave(t, false, c[i])
+			testLoadSave(t, c[i])
 			wg.Done()
 		}(i)
 	}
 	wg.Wait()
+	log.Print(tmp)
```
ineiti: `log.Print` is only for debugging, it should be removed afterwards.

service.go
```diff
@@ -228,13 +243,50 @@ func newServiceManager(c *Server, o *Overlay) *serviceManager {
 	return s
 }
 
+// openDb opens a database at `path`. It creates the database if it does not exist.
+func openDb(path string) (*bolt.DB, error) {
+	db, err := bolt.Open(path, 0600, nil)
```
ineiti: what happens if the path doesn't exist? Will it be created?

context.go
```diff
@@ -2,14 +2,13 @@ package onet
...
 	"errors"
-	"fmt"
-	"io/ioutil"
 	"os"
 	"os/user"
 	"path"
 	"runtime"
 	"sync"
 
+	bolt "github.com/coreos/bbolt"
```
ineiti: Is this the standard import for bbolt? Else using `bbolt.` is not too bad, neither. I prefer not overwriting the package name in general.

context.go
```diff
@@ -130,79 +129,61 @@ var testContextData = struct {
...
-		var err error
-		buf, err = ioutil.ReadFile(c.absFilename(id))
-		if err != nil {
-			return nil, err
-		}
+	bucketName := ServiceFactory.Name(c.ServiceID())
+	c.manager.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(bucketName))
+		v := b.Get([]byte(id))
+		buf = make([]byte, len(v))
```
jeffallen: Because of Boltdb's API, you are going to have this make and copy thing a lot. It is not really more readable, but it is more compact to do:
v:= b.Get([]byte(id))
return append([]byte(nil), v...)

It theoretically also informs the runtime that it could give you memory that has not been zeroed, because it knows for sure that it will be filling it before you get it back control. But I bet (a) there's no measurable perf impact from that, and (b) Go doesn't do it.

context.go
kc1212: This is what is used in all the bbolt example (https://github.com/coreos/bbolt). I'm guessing it's because bbolt is a fork of bolt, but to maintain compatibility, they rename the import to bolt.

context.go
```diff
@@ -130,79 +129,61 @@ var testContextData = struct {
...
 func (c *Context) DataAvailable(id string) bool {
-	if getContextDataPath() == "" {
-		testContextData.Lock()
-		_, ok := testContextData.service[c.absFilename(id)]
-		testContextData.Unlock()
-		return ok
-	}
-	_, err := os.Stat(c.absFilename(id))
-	return !os.IsNotExist(err)
+	available := false
```
jeffallen: Kind of hard to see why DataAvailable is such a useful part of the API. This work is already done, but you should never be afraid of saying, "instead of changing this code, I'm just going to remove it and adapt the 2 places that call it". For example here, Load() could be documented to return nil, nil when no data is available, and boom, you're done.

context.go
```diff
@@ -130,79 +129,61 @@ var testContextData = struct {
...
 }
 
-// absFilename returns the absolute path to load and save the configuration.
-// The file is chosen as "#{ServerIdentity.Public}_#{ServiceName}_#{id}.bin",
-// so no service and no server share the same file.
-func (c *Context) absFilename(id string) string {
-	pub, _ := c.ServerIdentity().Public.MarshalBinary()
-	return path.Join(getContextDataPath(), fmt.Sprintf("%x_%s_%s.bin", pub,
-		ServiceFactory.Name(c.ServiceID()), id))
+// GetDbAndBucket returns the DB handler and the bucket name of the service
```
jeffallen: End sentence with period, please.

context_test.go
```diff
@@ -20,60 +22,48 @@ type ContextData struct {
 }
 
 func TestContextSaveLoad(t *testing.T) {
-	setContextDataPath("")
+	tmp, err := ioutil.TempDir("", "conode")
```
jeffallen: check err please. that os.RemoveAll with an unknown tmp is scary!

context_test.go
jeffallen: remove probably? or t.Log?

service.go
kc1212: If the database does not exist, it will be created. If the parent path does not exist, e.g. `.local/share/conode`, then an error should be returned. Which I believe is fine because this function should only be called after `contextDataPath` is initialised. I can add a comment in there to clarify.

service.go
```diff
@@ -228,13 +243,50 @@ func newServiceManager(c *Server, o *Overlay) *serviceManager {
...
+
 // Process implements the Processor interface: service manager will relay
 // messages to the right Service.
 func (s *serviceManager) Process(env *network.Envelope) {
 	// will launch a go routine for that message
 	s.Dispatch(env)
 }
 
+// CloseDatabase closes the database. It also removes the database file if
+// contestDataPath is set to "".
```
jeffallen: s/contest/context/

context.go
kc1212: The Save function doesn't create the files anymore, so I removed this bit of the comment. But it's still relevant for `initContextDataPath()`, so I'll add it back there

context.go
kc1212: Good point. Actually, there are only two places where `DataAvailable` is used (in cothority)
```
identity/service.go
697:	if !s.DataAvailable("storage") {

pop/service/service.go
854:	if !s.DataAvailable("storage") {
```
In both of these cases `s.Load` immediately follows `s.DataAvailable`. I'd be clearer and simpler to remove this function, it'll also be closer to the functionality in boltdb.

context.go
kc1212: I don't think this will work. The callback needs to return an error type.

service.go
ineiti: OK, small comment is enough. I was not sure anymore if we correctly create the `.local/sahre/conode`-path...

context.go
ineiti: IIRC I used `DataAvailable` because I can store a `nil` data - and having `nil` data is different from having no data. But perhaps that's just my memory playing me tricks...

kc1212:
@jeffallen @ineiti hopefully the latest commit c8fba86 addressed all the issues

context_test.go
ineiti: Don't mix logging packages. If you think this might be useful, put something like
```
log.Lvl2("temp-directory is:", tmp)
```

kc1212: Forgot to delete that line, thanks!

service.go
```diff
@@ -209,12 +213,23 @@ const configFolder = "config"
...
-	s := &serviceManager{services, c, network.NewRoutineDispatcher()}
+	s := &serviceManager{services, c, db, network.NewRoutineDispatcher()}
 	ids := ServiceFactory.registeredServiceIDs()
 	for _, id := range ids {
 		name := ServiceFactory.Name(id)
 		log.Lvl3("Starting service", name)
+
+		err = createBucketForService(s.db, name)
+		if err != nil {
+			log.Panic("Failed to create bucket: " + err.Error())
```
jeffallen: Panic does not seem like the right choice.

kc1212: According to the documentation, `tx.CreateBucketIfNotExists` would only fail if the bucket name is empty or it is too long. As far as I understand, it's not like a user input and the programmer should have full control of the service name. If it is empty or too long, then it's a programmer error rather than an error that can be handled, hence we panic.

jeffallen: ok, got it.

PR #264, opened 2017-12-11
implement the counterio interface

ineiti:
Make it possible for services to measure their bandwidth in a simulation


ineiti:
Closed by #354

PR #265, opened 2017-12-11
implement the counterio interface

ineiti:
Make it possible for a service to measure bandwidth of operations during a simulation.

### DIFF
--- a/overlay.go
+++ b/overlay.go
@@ -327,6 +327,16 @@ func (o *Overlay) TreeNodeFromToken(t *Token) (*TreeNode, error) {
 	return tn, nil
 }
 
+// Rx() implements the CounterIO interface
+func (c *Overlay) Rx() uint64 {
+	return c.server.Rx()
+}
+
+// Tx() implements the CounterIO interface
+func (c *Overlay) Tx() uint64 {
+	return c.server.Tx()
+}
+
 func (o *Overlay) handleRequestTree(si *network.ServerIdentity, req *RequestTree, io MessageProxy) {
 	tid := req.TreeID
 	tree := o.Tree(tid)
--- a/overlay.go
+++ b/overlay.go
@@ -327,14 +327,14 @@ func (o *Overlay) TreeNodeFromToken(t *Token) (*TreeNode, error) {
 	return tn, nil
 }
 
-// Rx() implements the CounterIO interface
-func (c *Overlay) Rx() uint64 {
-	return c.server.Rx()
+// Rx implements the CounterIO interface
+func (o *Overlay) Rx() uint64 {
+	return o.server.Rx()
 }
 
-// Tx() implements the CounterIO interface
-func (c *Overlay) Tx() uint64 {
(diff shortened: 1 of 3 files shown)


overlay.go
```diff
@@ -327,6 +327,16 @@ func (o *Overlay) TreeNodeFromToken(t *Token) (*TreeNode, error) {
 	return tn, nil
 }
 
+// Rx() implements the CounterIO interface
+func (c *Overlay) Rx() uint64 {
```
nikkolasg: This is a bit misleading since the overlay's bandwidth may not be the same as the server's bandwidth. Service's messages don't necessarily pass through the overlay. 
Since the `Overlay` is marshalling the message itself, it may be reasonable to implement accurately this interface counting only the messages it marshals and unmarshals ?
**(Same comment for TreeNodeInstance)**

kc1212: I'm looking at implementing this feature. If I'm understanding this correctly, for `Overlay`, we need to modify the `Wrap/Unwrap` functions from the `MessageProxy` interface to keep the tx/rx values. For `TreeNodeInstance`, we need to modify `dispatchMsgToProtocol/dispatchMsgReader` to do the same. Does that sound correct?

kc1212:
There doesn't seem to be a straightforward solution from what I can see.

Overlay is something that routes messages to to the correct destinations, so it doesn't unmarshal everything, only enough to figure out the message type and what to do about it. For example, all the messages seem to go through the `MessageProxy`, but it only marshal/unmarshals `ProtocolMsg`, the other messages are left as is. Tx/rx is also difficult to implement for TreeNodeInstance because it eventually sends all the messages to the Overlay without doing any marshalling/unmarshalling.

I have the following proposal. Since everything eventually uses `Send/Receive` from the `network.Conn` interface and the messages are marshalled/unmarshalled there, why don't we add an additional return value, i.e. change from `Send(Message) error` to `Send(Message) (int64, error)`, where the int64 represents the number of bytes sent. That way, we can have the exact number of bytes sent/received propagate back to the caller and any objects in the return chain can decide whether to store the tx/rx. This is in fact similar to what the `net` package does from the standard library.

nikkolasg:
> That way, we can have the exact number of bytes sent/received propagate back to the caller and any objects in the return chain can decide whether to store the tx/rx. 
That sounds reasonable enough ! Altough do you think it's possible to make it so that the `Server.Rx()` is greater than the `Overlay.Rx()`,i.e. that `Overlay` modify the Tx accordingly to only what it has seen ? If it's not possible then it might not be a big deal either, but this needs to be clearly stated that the numbers are the one from the Server, just there to get those numbers in a different number of places... (which eventually we'll need something else to pass values amongst all components of onet but that's another story)

ineiti:
I think `Overlay.(T|R)x` is always the same as `Server.(T|R)x`. Which is not true for `TreeNodeInstance` and `Context`.

ineiti:
Closed by #354

PR #268, opened 2017-12-14
remove service_storage directory after each test

ineiti:
One liner to remove the storage directory between each integration test.


kc1212:
Regarding shell scripts, it seems we don't use a linter, locally or travis. My linter (shellcheck) tells me there are a lot of problems with this script.

ineiti:
Can you open an issue for the `shellcheck`? Then we can run it against all different shell-scripts we have - it complains a lot ;)

PR #270, opened 2017-12-14
Do not use base64 in toml files anymore.

jeffallen:
Closes onet#267

### DIFF
--- a/onet.go
+++ b/onet.go
@@ -18,4 +18,9 @@ ProtocolInstance.
 package onet
 
 // Version of onet.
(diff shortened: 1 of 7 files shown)


onet.go
```diff
@@ -18,4 +18,9 @@ ProtocolInstance.
 package onet
 
 // Version of onet.
-const Version = "1.2"
+const Version = "1.3"
```
ineiti: As it's breaking, I would not put it as a `v1`-version, but rather a `v2-alpha` or so.

PR #275, opened 2017-12-15
adding new function to context to create a new bucket

ineiti:
Instead of returning the existing bucket, change the method to create a new bucket, if it doesn't exist yet. It also makes sure that the names are conode-unique by prepending the name of the service to the request.

### DIFF
--- a/context.go
+++ b/context.go
@@ -171,7 +171,7 @@ func (c *Context) Load(key string) (interface{}, error) {
 // exists, by eventually creating it, and returns the created bucket name,
 // which is the servicename + the given name.
 func (c *Context) GetAdditionalBucket(name string) (*bolt.DB, string) {
-	fullName := c.bucketName + name
+	fullName := c.bucketName + "_" + name
 	err := c.manager.db.Update(func(tx *bolt.Tx) error {
 		_, err := tx.CreateBucketIfNotExists([]byte(fullName))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -166,10 +167,22 @@ func (c *Context) Load(key string) (interface{}, error) {
...
 
-// GetDbAndBucket returns the DB handler and the bucket name of the service.
-// The server should have created the database before calling this function.
-func (c *Context) GetDbAndBucket() (*bolt.DB, string) {
-	return c.manager.db, c.bucketName
+// GetAdditionalBucket makes sure that a bucket with the given name
+// exists, by eventually creating it, and returns the created bucket name,
+// which is the servicename + the given name.
+func (c *Context) GetAdditionalBucket(name string) (*bolt.DB, string) {
+	fullName := c.bucketName + name
```
kc1212: what happens if name is empty? I think maybe we need to at least log something as a warning.

context.go
```diff
@@ -166,10 +167,22 @@ func (c *Context) Load(key string) (interface{}, error) {
...
+	err := c.manager.db.Update(func(tx *bolt.Tx) error {
+		_, err := tx.CreateBucketIfNotExists([]byte(fullName))
+		if err != nil {
+			return fmt.Errorf("create bucket: %s", err)
+		}
+		return nil
+	})
+	if err != nil {
```
kc1212: this error is never reported

PR #281, opened 2017-12-21
Unlock dispatcher before processing message

### LINKED ISSUE #277: Calling RegisterProcessorFunc can deadlock
If you call RegisterProcessorFunc from inside a message processor, you can attempt to double lock the dispatcher.

Reported by Raphael Dunant.

jeffallen:
If the processor for the message wants to call back into
the dispatcher for something, it will double lock, and deadlock.
Release the lock after the map read is done.

Closes #277


ineiti:
Will this also have to go in `master`?

jeffallen:
It is already on master. The only way to commit to v1.2 now is cherry picking.

PR #286, opened 2017-12-28
Ignore errors if fewer than threshold

kc1212:
Not sure about the function name, any suggestions for a better one?


kc1212:
Perhaps change the name to `SendToChildrenInParallel_IgnoreTErrs`, but maybe too long?

treenode.go
```diff
@@ -646,6 +646,20 @@ func (n *TreeNodeInstance) SendToChildren(msg interface{}) error {
 // If the underlying node is a leaf node this function does
 // nothing.
 func (n *TreeNodeInstance) SendToChildrenInParallel(msg interface{}) error {
+	return collectErrors("Error while sending to %s: %s\n", n.sendToChildrenInParallel(msg))
+}
+
+// SendToChildrenInParallelIgnoreErrs is the same as SendToChildrenInParallel,
+// except it ignores errors if there are fewer than `t` of them.
+func (n *TreeNodeInstance) SendToChildrenInParallelIgnoreErrs(msg interface{}, t int) error {
```
jeffallen: SendToChildrenInParallel is called in so few places, maybe it would be better for you to just refactor it now to return []error. Then where you need to check if len(errors < X) you can.

jeffallen: The callers are:
./cothority/bftcosi/bftcosi.go:	return bft.SendToChildrenInParallel(&ann)
./cothority/bftcosi/bftcosi.go:	return bft.SendToChildrenInParallel(&ch)
./cothority/bftcosi/bftcosi.go:	return bft.SendToChildrenInParallel(&ch)

kc1212: Fixed now

treenode.go
```diff
@@ -659,14 +659,14 @@ func (n *TreeNodeInstance) SendToChildrenInParallel(msg interface{}) error {
 		go func(n2 *TreeNode) {
 			if err := n.SendTo(n2, msg); err != nil {
 				eMut.Lock()
-				errs = append(errs, collectedErrors{name, err})
+				errs = append(errs, errors.New(name+": "+err.Error()))
```
jeffallen: You don't need to do this right now, but it is better to keep error info in machine readable format, instead of collapsing it into strings. So instead of errors.New, you'd make a new struct called NodeError{ Node *TreeNode, Error error } and append those onto errs. Type NodeError would need to implement method "Error() string" in order to implement error.

PR #287, opened 2018-01-08
Change to a versioned import path for go.uuid.

jeffallen:


jeffallen:
go.uuid is still under active dev after ~12 months quiet. I would rather just get onto a stable version. Also, the dev is making breaking changes in master, that's his right. We need to get off of master so he can hack there without breaking us.

PR #289, opened 2018-01-09
Change to use kyber.v1

jeffallen:
Also: simplify travis configs and de-flake a test.

### DIFF
--- a/service_test.go
+++ b/service_test.go
@@ -687,7 +687,7 @@ func waitOrFatal(ch chan bool, t *testing.T) {
 	select {
 	case _ = <-ch:
 		return
-	case <-time.After(500 * time.Millisecond):
+	case <-time.After(1500 * time.Millisecond):
 		t.Fatal("Waited too long")
 	}
 }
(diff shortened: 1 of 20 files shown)


service_test.go
```diff
@@ -687,7 +687,7 @@ func waitOrFatal(ch chan bool, t *testing.T) {
 	select {
 	case _ = <-ch:
 		return
-	case <-time.After(500 * time.Millisecond):
+	case <-time.After(1500 * time.Millisecond):
```
kc1212: @jeffallen why this change? is it because travis is slow?

ineiti:
I think I missed something here - I thought we're going for 

kyber.master <- onet.master <- cothority.master

and now onet.master should depend on kyber.v1?

jeffallen:
I'm not certain what we want, this is an experiment to check that onet/cothority work right with kyber.v1. When we freeze onet and cothority (at v2, I think?) then they should be frozen to kyber.v1 as well, I guess. This was not ready for merge. that was an accident. I've moved it back to WIP.

service_test.go
jeffallen: Maybe? It was an experiment. I'm not satisfied with it, but it is hard to see what else to do there.

ineiti:
Thanks for the explanation - makes sense now!

jeffallen:
This was an experiment. Closing for now.

PR #290, opened 2018-01-10
Fix integration tests

jeffallen:


app/libtest.sh
```diff
@@ -256,7 +258,20 @@ setupConode(){
 runCoBG(){
     for nb in "$@"; do
     	testOut "starting conode-server #$nb"
-    	( ./conode -d $DBG_SRV -c co$nb/private.toml | tee $COLOG$nb.log & )
+    	(
+	    rm -f "$COLOG$nb.log.dead"
+	    ./conode -d $DBG_SRV -c co$nb/private.toml server 2>&1 | tee "$COLOG$nb.log"
+	    touch "$COLOG$nb.log.dead"
```
kc1212: would this leave extra files after the test completes?

app/libtest.sh
```diff
@@ -256,7 +258,20 @@ setupConode(){
 runCoBG(){
     for nb in "$@"; do
     	testOut "starting conode-server #$nb"
-    	( ./conode -d $DBG_SRV -c co$nb/private.toml | tee $COLOG$nb.log & )
+    	(
```
jeffallen: The changes in this area are to make the tests explain themselves better when something goes wrong. Instead of just hoping the server is up and running, we check to see if it died on start.

overlay_test.go
```diff
@@ -25,10 +25,7 @@ func (po *ProtocolOverlay) Start() error {
 
 func (po *ProtocolOverlay) Dispatch() error {
 	if po.failDispatch {
-		go func() {
-			// hopefully we write to channel after `Dispatch` returns
-			po.failChan <- true
-		}()
+		po.failChan <- true
```
kc1212: could you add a comment here to stay ideally this should happen after `return errors.New("Dispatch failed")`? otherwise, if the runtime decides to execute in the following order:
```
1, po.failChain <- true
2, assert.Contains(t, log.GetStdOut()+log.GetStdErr(), "Dispatch failed")
3, return errors.New("Dispatch failed")
```
the test would fail (although i have no idea if it's possible for the runtime to do something like that)

app/commands.go
```diff
@@ -1,96 +0,0 @@
-package app
```
kc1212: why are these files removed?

jeffallen: They duplicated the code in conode.go which is used now instead.

app/libtest.sh
jeffallen: the current working directory is the build dir, so it is cleaned up (or not, if -nt is given as an argument)

PR #291, opened 2018-01-10
Fix the build for Windows.

jeffallen:


kc1212:
Travis doesn't do windows right? I also don't have a windows machine to test the fix. How did this problem come up?

jeffallen:
I used Go cross compilation to check windows building of the fd_limit stuff, and the build failed because of onet. Some of the students have Windows laptops and thus use Windows.

We can set up travis matrix builds to try building with many different configs, including cross compilation to Windows and with vartime.

PR #296, opened 2018-01-12
Support choosing the suite from the config file

jeffallen:
Make it possible to choose the suite from the config file.

Also: Various small fixes to make testing this branch work reliably.


tree.go
```diff
@@ -388,7 +388,9 @@ func NewRoster(ids []*network.ServerIdentity) *Roster {
 		agg := ids[0].Public.Clone()
 		agg.Null()
 		for _, e := range ids {
-			agg = agg.Add(agg, e.Public)
+			if e.Public != nil {
+				agg = agg.Add(agg, e.Public)
+			}
```
kc1212: can `e.Public` be nil?  My guess is that it can't because `ids` should have actual content, right? Maybe we need to return a nil roster in this case.

network/local.go
```diff
@@ -41,6 +41,8 @@ type LocalManager struct {
 	// close on this channel indicates that connection retries should stop
 	stopping chan bool
 	stopOnce sync.Once
+	// a waitgroup to check that all serving goroutines are done
```
ineiti: Does that change belong into the current PR?

jeffallen: Without this, the builds were not passing, so no it's not related, but yes it's needed.

ineiti:
Please provide a description and fix the `.travis.yml`-error

tree.go
```diff
@@ -374,13 +374,20 @@ func (roID RosterID) IsNil() bool {
 // RosterTypeID of Roster message as registered in network
 var RosterTypeID = network.RegisterMessage(Roster{})
 
-// NewRoster creates a new ServerIdentity from a list of entities. It also
+// NewRoster creates a new roster from a list of entities. It also
 // adds a UUID which is randomly chosen.
 func NewRoster(ids []*network.ServerIdentity) *Roster {
+	// Don't allow a crash if things are not as expected.
+	if len(ids) < 0 || ids[0].Public == nil {
```
kc1212: @jeffallen sorry, didn't notice this when I first reviewed it. You want `len(ids) < 1`, right? Can you change it in a future commit?

jeffallen: Fixing with issue274.

PR #304, opened 2018-01-23
adding RegisterChannelsLength convenience method

ineiti:
I wanted to do this a long time already. Now we have all four:
- `RegisterChannel`
- `RegisterChannels`
- `RegisterChannelLength`
- `RegisterChannelsLength`


jeffallen:
YAGNI. But I'll merge it anway, because you're a nice enough fellow.

PR #306, opened 2018-01-24
Make DB files in temporary directories during tests

### LINKED ISSUE #274: Database-storage
I would like the database-storage for tests to be in a temporary directory. When I test and the test fails with a `Fatal`, the db-files are still in my directory, which clutters things.

jeffallen:
Creation and deletion of the temp directory is managed
by serviceManager. There is one serviceManager per *onet.Server.

The ways to create *onet.Server's had to be constrained to
either test-mode, which ensures the db is nuked, or live mode
which does not delete the database. Only servers created via
the methods on *onet.LocalTest delete the db. The other ways
to create servers leave them in live mode, so there can be no accidental
DB deletion.

Simulation had to be modified to use the environment variable
and later cleanup.

Tests had to be modified to use local.NewLocalTest if they were not
already.

Fixes #274.


service.go
```diff
@@ -205,21 +206,28 @@ type serviceManager struct {
 	server *Server
 	// a bbolt database for all services
 	db *bolt.DB
+	// where the database should be
+	where string
```
kc1212: rather than where, it's clearer to say something like `dbDir`

service.go
```diff
@@ -269,7 +286,7 @@ func (s *serviceManager) Process(env *network.Envelope) {
 }
 
 // closeDatabase closes the database.
-// It also removes the database file if contextDataPath is set to "".
+// It also removes the database file if the path is not default (i.e. testing config)
```
kc1212: Does this mean we should never use non-default db-directory when writing prod code?

jeffallen: Yes, just as you would have never called setContextPath("") in production code.

kc1212: I feel this approach will cause some confusing (but no more than before) as `NewServer` is a public API, whoever is calling it might set the directory to something other than `""`. Do you think an explicit flag is better, something like `cleanOnClose bool` in `newServiceManager`? I remember Linus saying something about setting a flag to indicate whether the server is in a test or not.

service.go
```diff
@@ -204,22 +205,33 @@ type serviceManager struct {
...
+func newServiceManager(svr *Server, o *Overlay, dbPath string) *serviceManager {
+	services := make(map[ServiceID]Service)
+	s := &serviceManager{
+		services:   services,
+		server:     svr,
+		dbPath:     dbPath,
+		Dispatcher: network.NewRoutineDispatcher(),
+	}
+	if dbPath == "" {
+		s.delDb = true
```
kc1212: I'm nitpicking a bit - if we already have `delDb`, we don't need to check for `dbPath` anymore, right? `newServiceManager` and `newService` would take an additional parameter for `delDb`.

server.go
```diff
@@ -41,18 +40,22 @@ type Server struct {
 }
 
 // NewServer returns a fresh Server tied to a given Router.
-func NewServer(r *network.Router, pkey kyber.Scalar, s network.Suite) *Server {
+// If where is "", the server will write its database to the default
+// location. If where is != "", it is considered a temp dir, and the
+// DB is deleted on close.
+func newServer(where string, r *network.Router, pkey kyber.Scalar, s network.Suite) *Server {
```
kc1212: `s/where/delDb`

jeffallen: I missed that one before, pushign a new change to fix it.

service.go
kc1212: ops, `newServer`, not `newService`

jeffallen:
Wait, this is not ready to check in. I am getting left over *.db files in ~/Library/Conode now. Arrgh.

ineiti:
Can you give some more context in the description how you fixed this?

kc1212:
@jeffallen please fix conflicts

ineiti:
Did you try this out with the current cothority?

jeffallen:
There's something not right in cothority, debugging...

jeffallen:
Ok, as soon as this is merged, we need to merge dedis/cothority#1009 which has a fix to make it work with this.

local.go
```diff
@@ -164,6 +173,12 @@ func (l *LocalTest) GenRosterFromHost(servers ...*Server) *Roster {
 	return list
 }
 
+func (l *LocalTest) panicClosed() {
```
kc1212: I don't really like this because whoever is writing a new function will need to remember to call `panicClosed`. However, I can't think of any other way to do it and it's not a huge problem if it's not called, it'll just take longer to debug.

Issue #51, opened 2017-01-06
Error in TreeNodeInstance.DispatchChannel


ineiti:
Once upon a while there was an error in DispatchChannel, but it only showed up occasionally on Travis. So there is some code waiting there to show something, just in case it happens again...

kc1212:
There's still code regarding this issue here https://github.com/dedis/onet/blob/5bce6550583377efd53cd4080e0cc18984c2ffd3/treenode.go#L373, can they be removed?

PR #307, opened 2018-01-24
Remove calls to log.Stack that are no longer needed

### LINKED ISSUE #301: remove calls to log.Stack
Taking an entire stack dump is expensive.

Doing it on every call to TreeNodeInstance.SendToParent is really not ok.

Remove this call. Audit others and remove them when they are in logs like this.

jeffallen:
Fixes #301, #51.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,20 @@
 language: go
 
 install:
-  - go get -t ./...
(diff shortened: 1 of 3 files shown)


jeffallen:
I added a second commit here that tests a new system to start builds on other repos. Let's see how it goes...

PR #313, opened 2018-01-29
fixing a bug in LocalTest cleanup

ineiti:
The `LocalTest.Stop` function was misbehaving in that it supposed all connections were closed before it got called, which is not always the case.

This PR also changes the way `local.CloseAll` behaves: if `go test` is called without `-v`, `local.CloseAll` reroutes all logging to `/dev/null` while closing, as there might be quite some error messages due to late packets arriving.

### DIFF
--- a/network/local.go
+++ b/network/local.go
@@ -40,7 +40,7 @@ type LocalManager struct {
 	counter uint64
 	// close on this channel indicates that connection retries should stop
 	stopping chan bool
-	stopOnce sync.Once
+	stopped  bool
 	// a waitgroup to check that all serving goroutines are done
 	wg sync.WaitGroup
 }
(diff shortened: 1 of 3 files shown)


network/local.go
```diff
@@ -176,7 +179,14 @@ func (lm *LocalManager) count() int {
 // Stop tells any connections that are sleeping on retry
 // to stop sleeping and return an error.
 func (lm *LocalManager) Stop() {
-	lm.stopOnce.Do(func() { close(lm.stopping) })
+	lm.Lock()
```
jeffallen: stopOnce was needed before due to panic on "closing a closed channel" for the close of lm.stopping. What changed to not need this?

ineiti: It seems that we call `Stop` only once now. What test was triggering the error?

jeffallen: Don't recall. Let's check this in and come back to it if we need to.

PR #318, opened 2018-01-30
Change timeout integers to time.Duration

### LINKED ISSUE #315: simul/manage Timeout is int not time.Duration
```
// PrepareCount is sent so that every node can contact the root to say
// the counting is still going on.
type PrepareCount struct {
	Timeout int
}
```

jeffallen:
Fixes #315.

### DIFF
--- a/simul/manage/count.go
+++ b/simul/manage/count.go
@@ -71,7 +71,7 @@ func NewCount(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
 	p := &ProtocolCount{
 		TreeNodeInstance: n,
 		Quit:             make(chan bool),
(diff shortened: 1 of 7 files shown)


simul/manage/count.go
```diff
@@ -71,7 +71,7 @@ func NewCount(n *onet.TreeNodeInstance) (onet.ProtocolInstance, error) {
 	p := &ProtocolCount{
 		TreeNodeInstance: n,
 		Quit:             make(chan bool),
-		timeout:          1024,
+		timeout:          1024 * time.Millisecond,
```
ineiti: git unblame ineiti: s/1024/1000/

PR #321, opened 2018-02-01
using localhost in libtest.sh

ineiti:
Replacing `127.0.0.1` with `localhost` in libtest.sh

### DIFF
--- a/service.go
+++ b/service.go
@@ -205,9 +205,8 @@ type serviceManager struct {
 	// the onet host
 	server *Server
 	// a bbolt database for all services
-	db      *bolt.DB
-	dbPath  string
-	dbMutex sync.Mutex
+	db     *bolt.DB
+	dbPath string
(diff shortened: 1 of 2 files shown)


service.go
```diff
@@ -286,6 +287,8 @@ func (s *serviceManager) Process(env *network.Envelope) {
 // closeDatabase closes the database.
 // It also removes the database file if the path is not default (i.e. testing config)
 func (s *serviceManager) closeDatabase() error {
+	s.dbMutex.Lock()
```
jeffallen: This seems like it is only protecting the write of "s.db = nil" below. How about if we just remove that, and leave db set. Other goroutines that try to act on s.db will then get some kind of "error already closed" error from boltdb: https://godoc.org/github.com/boltdb/bolt#ErrDatabaseNotOpen

PR #324, opened 2018-02-02
Make it easier to debug when RunWait is set wrong

### LINKED ISSUE #323: onet simulation RunWait
With the change of having real `time.Duration` instead of `int`, `RunWait` and `ExperimentWait` in the `simulation.toml`-files don't behave as expected.

Need to change:
- either print an error if `RunWait` or `ExperimentWait` misses time-unit
- or suppose seconds if `RunWait` or `ExperimentWait` are integers

Files (at least):
simul/build.go
simul/platform/platform.go

jeffallen:
Also: Fix the type of RunWait for deterlab.

Fixes #323.

### DIFF
--- a/simul/build.go
+++ b/simul/build.go
@@ -81,7 +81,10 @@ func startBuild() {
 		} else {
 			logname := strings.Replace(filepath.Base(simulation), ".toml", "", 1)
 			testsDone := make(chan bool)
-			timeout := getExperimentWait(runconfigs)
+			timeout, err := getExperimentWait(runconfigs)
+			if err != nil {
+				log.Fatal("ExperimentWait:", err)
+			}
 			go func() {
 				RunTests(logname, runconfigs)
 				testsDone <- true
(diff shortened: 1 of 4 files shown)


ineiti:
There are three more places to take into account for, one for sure:
- `simul/README.md`
and two to decide:
- `simul/platform/mininet.go`
- `simul/platform/deterlab.go`

simul/build.go
```diff
@@ -311,29 +318,42 @@ func getStartStop(rcs int) (int, int) {
...
+		return 0, err
 	}
+
+	// Otherwise calculate a useful default.
 	wait := 0 * time.Second
 	for _, rc := range rcs {
-		wait += getRunWait(rc)
+		w, err := getRunWait(rc)
+		if err != nil {
+			log.Fatal("RunWait:", err)
```
kc1212: we can return an error here instand

PR #328, opened 2018-02-06
Count bytes more correctly

### LINKED ISSUE #292: tcp sendRaw does not call updateTx on short write
If sendRaw exits via handleError because c.conn.Write fails, then updateTx is not called.

jeffallen:
Count all the bytes, and also count them if there's an error.
Also: defer costs a memory allocation, don't use them when not needed.

Fixes #292.

### DIFF
--- a/network/router_test.go
+++ b/network/router_test.go
@@ -196,6 +196,7 @@ func testRouterAutoConnection(t *testing.T, fac routerFactory) {
 	require.NotNil(t, h21)
 	assert.Nil(t, h21.Close())
 	time.Sleep(100 * time.Millisecond)
+
 	err = h1.Send(h2.ServerIdentity, &SimpleMessage{12})
 	require.Nil(t, err)
 	<-proc.relay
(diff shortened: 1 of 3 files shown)


network/router_test.go
```diff
@@ -204,8 +205,10 @@ func testRouterAutoConnection(t *testing.T, fac routerFactory) {
 		t.Fatal("Should be able to stop h2")
 	}
 	err = h1.Send(h2.ServerIdentity, &SimpleMessage{12})
+	if err == nil {
+		t.Error("h1 let us send still")
+	}
 	require.NotNil(t, err)
```
ineiti: Why not
```
require.NotNil(t, err, "h1 let us send still")
```
?

jeffallen: It was left over from debugging. I do not like how require prints its errors, because Emacs cannot jump to the line. Can we just agree that sometimes calling t.Error(f) is ok?

ineiti: OK for me.

PR #330, opened 2018-02-08
Move suite to first argument

kc1212:
I couple more that I missed in https://github.com/dedis/onet/pull/322


kc1212:
@ineiti no problem, just made a commit for it

PR #333, opened 2018-02-09
Implement TLS between conodes

jeffallen:


network/address.go
```diff
@@ -196,6 +198,10 @@ func (a Address) Valid() bool {
 		return false
 	}
 
+	// If hostname is missing then they mean "all IPs", and that's valid.
+	if len(ip) == 0 {
+		return true
+	}
```
kc1212: Can't seem to make a comment on unchanged lines. But the documentation for `Valid()` seem to be wrong. The example should be `tls://192.168.1.10:5678` and not `tls:192.168.1.10:5678`.

network/tcp.go
```diff
@@ -111,7 +107,7 @@ func (c *TCPConn) receiveRaw() ([]byte, error) {
 		return nil, handleError(err)
 	}
 	if total > MaxPacketSize {
-		return nil, errors.New(c.endpoint.String() + " sends too big packet")
+		return nil, fmt.Errorf(c.conn.RemoteAddr().String()+" sends too big packet %v>%v", total, MaxPacketSize)
```
kc1212: nitpick: you're already using `Errorf`, why not just do 
```(%s sends too big packet ...", c.conn.RemoteAddr().String(), ...)```

File: network/tls.go [hunk 3]
```diff
@@ -0,0 +1,415 @@
...
+// key for the self-signed TLS cert) also holds the conode's private key.
+// We do this by hashing a nonce provided by the peer (in order to prove that this
+// is a fresh challenge response) and the ASN.1 encoded CommonName of the certificate.
+// The CN is always the hex-encoded form of the conode's public key.
+//
+// Because each side needs a nonce which is controlled by the opposite party in the
+// mutual authentication, but TLS does not support sending application data before the handshake,
+// we need to find places in the normal TLS 1.2 handshake where we can "tunnel" the nonce
+// through. On the TLS client side, we use ClientHelloInfo.ServerName. On the
+// the TLS server side, we use the.
```
kc1212: "we use the"?

network/tls.go
```diff
@@ -0,0 +1,415 @@
...
+// for the moment we are not targeting other languages than Go on the conode/conode
+// communication channel.
+
+// TODO: Websockets.
+// All of this is completely unrelated to HTTPS security on the websocket side. For
+// that, we will implement an opt-in Let's Encrypt client in websocket.go.
+
+// certMaker holds the data necessary to make a certificate on the fly
+// and give it to crypto/tls via the GetCertificate and
+// GetClientCertificate callbacks.
```
kc1212: `GetCertificate` and `GetClientCertificate` are private now

network/tls.go
```diff
@@ -0,0 +1,415 @@
...
+// that is created on server boot and stored in RAM. Because the certificates are
+// self-signed, there is no need to pay a CA, nor to load the private key
+// or certificate from disk.
+//
+// In order to achieve #3, we include an extension in the certificate, which
+// proves that the same entity that holds the TLS private key (i.e. the signing
+// key for the self-signed TLS cert) also holds the conode's private key.
+// We do this by hashing a nonce provided by the peer (in order to prove that this
+// is a fresh challenge response) and the ASN.1 encoded CommonName of the certificate.
+// The CN is always the hex-encoded form of the conode's public key.
```
kc1212: does this mean we cannot use TLS in the normal way? i.e. proof that we actually own the domain name

network/tls.go
```diff
@@ -0,0 +1,415 @@
...
+// that, we will implement an opt-in Let's Encrypt client in websocket.go.
+
+// certMaker holds the data necessary to make a certificate on the fly
+// and give it to crypto/tls via the GetCertificate and
+// GetClientCertificate callbacks.
+type certMaker struct {
+	si      *ServerIdentity
+	suite   Suite
+	subj    pkix.Name
+	subjDer []byte
```
kc1212: could you describe what some of these fields are? I don't know what `subjDer` is.

network/tls.go
```diff
@@ -0,0 +1,415 @@
...
+		PrivateKey:  cm.k,
+		Certificate: [][]byte{cDer},
+		Leaf:        certs[0],
+	}, nil
+}
+
+// See https://github.com/dedis/Coding/tree/master/mib/cothority.mib
+var oidDedisSig = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 51281, 1, 1}
+
+func isDedisSig(in asn1.ObjectIdentifier) bool {
```
kc1212: There's an `.Equal` function https://golang.org/pkg/encoding/asn1/#ObjectIdentifier.Equal

network/tls.go
```diff
@@ -0,0 +1,415 @@
...
+	// will be able to easily do so with their own x509 + kyber implementation.
+	buf := bytes.NewBuffer(nonce)
+	buf.Write(cm.subjDer)
+	sig, err := schnorr.Sign(cm.suite, cm.si.private, buf.Bytes())
+	if err != nil {
+		return nil, err
+	}
+
+	// The serial number is non-security critical how we are using it here,
+	// so just make it a big random number.
```
kc1212: Is this comment still up to date? It's using the random number generator from kyber, which I imagine is secure.

network/router.go
```diff
@@ -397,13 +402,35 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
...
 	}
+
 	// Set the ServerIdentity for this connection
 	dst := nm.Msg.(*ServerIdentity)
 
-	if err != nil {
-		return nil, err
+	// See if we have a cryptographically proven pubkey for this peer. If so,
+	// check it against dst.Public.
+	if tcpConn, ok := c.(*TCPConn); ok {
```
nikkolasg: We've discussed that already with @jeffallen but I wanted to write it down for public scrutiny:
I'm a little bit hesitant about "breaking" the interface in the router such as what's done here. The router should not care, should not know what precise implementation of the connection he is using. Of course the problem currently is that the router is doing the "authentication" (well, it's not even that but let's assume), by requiring an exchange of server identity on top of the connection. So there is a duplicate effort here with the TLS connection which already provides that, and has two consequences: (1) one more round trip for a ServerIdentity that we should not really need anymore (I guess it could be embedded in the certificate right ?) (2) it breaks the nice layer between a `Router` and its underlying transport. 
In order to solve that, that would maybe need some further restructuring like putting the authentication part one layer below: in the transport layer. The transport layer would then give a tuple `(ServerIdentity, Conn)` back to the router. For TLS connection, it seems possible to do that without too much hassle. For TCP connection, we can simply put the current authentication method down in the TCP stack, so when a new TCP connection is created, one expects a ServerIdentity exchange to happen, and then pass the tuple up to the router when ready. Same for Local connection.

Anyway, I'm not strongly suggesting that the current PR should not be merged, but merely proposing what I see as a smoother and nicer way to integrate real authentication in this stack.

network/tcp.go
```diff
@@ -379,38 +378,37 @@ func (t *TCPListener) Listening() bool {
...
 type TCPHost struct {
-	addr  Address
 	suite Suite
+	sid   *ServerIdentity
 	*TCPListener
 }
 
 // NewTCPHost returns a new Host using TCP connection based type.
-func NewTCPHost(addr Address, s Suite) (*TCPHost, error) {
+func NewTCPHost(sid *ServerIdentity, s Suite) (*TCPHost, error) {
```
nikkolasg: Why not making a `NewTLSHost` ?
It seems cleaner to separate the two, and maybe write a `deprecated plain TCP connection` on the log or stg ?

network/tcp.go
```diff
@@ -379,38 +378,37 @@ func (t *TCPListener) Listening() bool {
...
-		addr:  addr,
 		suite: s,
+		sid:   sid,
 	}
 	var err error
-	h.TCPListener, err = NewTCPListener(addr, s)
+	if sid.Address.ConnType() == TLS {
+		h.TCPListener, err = NewTLSListener(sid, s)
+	} else {
+		h.TCPListener, err = NewTCPListener(sid.Address, s)
```
nikkolasg: Maybe write a `deprecated plain TCP connection` on the log or stg ?

network/tls.go
jeffallen: I was talking about the callbacks in the tls.Config structure. I fixed the comment to be more precise.

network/tls.go
jeffallen: Yes, for conode/conode communication, our definition of "a valid certificate: is nonstandard. We can use CN's related to our domain name on the websocket side.

network/tls.go
jeffallen: DER = https://en.wikipedia.org/wiki/X.690#DER_encoding 🤢

network/tcp.go
jeffallen: They are not deprecated. If TCP was the right choice for an application, we can support it still.

network/tcp.go
jeffallen: Honestly, this is the first thing that worked, and I was too tired to change it more. Assuming TCP happened all over the tests.

network/tcp.go
nikkolasg: Should it not be deprecated ?  Since you implemented encrypted communication by keeping the same API and the same deployement workflow, we should encourage the use of TLS and start deprecating less secure communication transport.

kc1212:
The build is failing. Is this a new flaky test? I haven't seen it before. https://travis-ci.org/dedis/onet/builds/340938361

network/router.go
ineiti: Looks like a good proposition for a further issue.

ineiti: Opened #338

network/tcp.go
ineiti: This is the only thing that I still _would_ still change - a `NewTLSHost` makes sense for me.

network/tcp.go
ineiti: I think we should keep the TCP connections around - if it's just for easier debugging. So -1 for deprecation warning.

ineiti:
I also have it failing on my local machine in the same way. So I think Travis is right (for once ;)

network/tcp.go
jeffallen: I have looked into this some. If one really wants to "clean this up", and I don't see what it messy and needs cleaning, then you'd want to kill New*Host entirely and have only NewHost(...) (Host, error). I've started on that, but then tests do not compile because they are written assuming that they can muck around inside the returned *TCPHost.

TLS happens inside of TCP. I do not see the problem of having TCP with TLS and PlainTCP (i.e. TCP without TLS) sharing the same data structure.

PR #336, opened 2018-02-13
Tune logging for production use

jeffallen:
Tune logging so that auditing and errors go out at level 2.

Also: Add an HTTP responder for "/ok" so that HTTP load balancers can detect if a conode is alive.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -42,6 +42,11 @@ func NewWebSocket(si *network.ServerIdentity) *WebSocket {
 	webHost, err := getWebAddress(si, true)
 	log.ErrFatal(err)
 	w.mux = http.NewServeMux()
+	w.mux.HandleFunc("/ok", func(w http.ResponseWriter, r *http.Request) {
+		log.Lvl2("ok?", r.RemoteAddr)
+		ok := []byte("ok\n")
+		w.Write(ok)
+	})
 	w.server = &graceful.Server{
 		Timeout: 100 * time.Millisecond,
 		Server: &http.Server{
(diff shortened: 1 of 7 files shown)


websocket.go
```diff
@@ -124,13 +138,16 @@ func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			err = rerr
 			break
 		}
+		rx += len(buf)
+		n++
```
kc1212: Ideally, we only want to do these if logging is needed. But it might be too much effort for very little gain.

websocket.go
```diff
@@ -42,6 +42,11 @@ func NewWebSocket(si *network.ServerIdentity) *WebSocket {
 	webHost, err := getWebAddress(si, true)
 	log.ErrFatal(err)
 	w.mux = http.NewServeMux()
+	w.mux.HandleFunc("/ok", func(w http.ResponseWriter, r *http.Request) {
+		log.Lvl2("ok?", r.RemoteAddr)
+		ok := []byte("ok\n")
+		w.Write(ok)
+	})
```
kc1212: I don't know much about websockets, so let me if what I'm not making sense. Is this a standard thing to do? If it's not, do we need to write somewhere that prevents the user from making a service called "ok" to avoid conflicting handlers?

ineiti: Sounds good to have a fool's guard in `registerService`.

websocket.go
ineiti: I think the `if` will take more time than the add here ;)

kc1212: yeah, `if` won't be useful. I was thinking more of a decorator pattern, but that's going to change everything.

PR #337, opened 2018-02-13
Give ProcessClientRequest the HTTP Request

### LINKED ISSUE #331: ProcessClientRequest does not have access to any HTTP client info
If a handler wanted to limit access to itself based on the origin IP address of the websocket client, it would not be able to. We need to pass something from the websocket or the HTTP request onwards to the service processor, so that when we make other service processors (i.e. like one that can do authentication/authorization) then we'll have all the info we could want.

The same is likely true for the conode/conode communication side of things. There, we should be able to find the ServicerIdentity of the peer (and know that it was proven via TLS).

jeffallen:
So that it can make decisions based on who the
remote client is, or what the state of TLS is, etc.

Fixes #331.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -2,6 +2,7 @@ package onet
 
 import (
 	"errors"
+	"net/http"
 	"reflect"
 
 	"strings"
@@ -113,7 +114,7 @@ func (p *ServiceProcessor) NewProtocol(tn *TreeNodeInstance, conf *GenericConfig
 // ProcessClientRequest takes a request from a websocket client, calculates the reply
 // and sends it back. It uses the path to find the appropriate handler-
 // function. It implements the Server interface.
-func (p *ServiceProcessor) ProcessClientRequest(path string, buf []byte) ([]byte, error) {
+func (p *ServiceProcessor) ProcessClientRequest(req *http.Request, path string, buf []byte) ([]byte, error) {
 	mh, ok := p.handlers[path]
 	reply, err := func() (interface{}, error) {
 		if !ok {
--- a/service.go
+++ b/service.go
@@ -3,6 +3,7 @@ package onet
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"path"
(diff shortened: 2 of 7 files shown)


processor.go
```diff
@@ -113,7 +114,7 @@ func (p *ServiceProcessor) NewProtocol(tn *TreeNodeInstance, conf *GenericConfig
 // ProcessClientRequest takes a request from a websocket client, calculates the reply
 // and sends it back. It uses the path to find the appropriate handler-
 // function. It implements the Server interface.
-func (p *ServiceProcessor) ProcessClientRequest(path string, buf []byte) ([]byte, error) {
+func (p *ServiceProcessor) ProcessClientRequest(req *http.Request, path string, buf []byte) ([]byte, error) {
```
nikkolasg: Feels like `path` could and perhaps should be removed ? Redundand information.

service.go
```diff
@@ -37,7 +38,7 @@ type Service interface {
 	// service. It returns a message that will be sent back to the
 	// client. The returned error will be formatted as a websocket
 	// error code 4000, using the string form of the error as the message.
-	ProcessClientRequest(handler string, msg []byte) (reply []byte, err error)
+	ProcessClientRequest(req *http.Request, handler string, msg []byte) (reply []byte, err error)
```
ineiti: Copying @nikkolasg question: shouldn't we remove handler here?

processor.go
jeffallen: No, because the path has already been trimmed by the caller (to trim the service name) off the rfront. That would make all implementors of ProcessClientRequest responsible for doing this trimming, and maybe they don't even know the service name that they have been called for.

nikkolasg: I agree the information has been pre-processed but still if you look at it from a fresh point of view, you just see two arguments that are redundant. You can translate that to "I don't really know which one I should use if I want to code a new service." 
What do you think about having a wrapper around `http.Request` that can deliver the path with a simple method ?

PR #348, opened 2018-02-26
Remove unused context.Dispatcher

### LINKED ISSUE #230: Double implementation of RegisterProcessorFunc
The `Context`-structure from:

```
// Context represents the methods that are available to a service.
type Context struct {
	overlay   *Overlay
	server    *Server
	serviceID ServiceID
	manager   *serviceManager
	network.Dispatcher
}
```

has an implementation of the method `RegisterProcessor`. Once it is embedded as interface and set here:

```
func newContext(c *Server, o *Overlay, servID ServiceID, manager *serviceManager) *Context {
	return &Context{
		overlay:    o,
		server:     c,
		serviceID:  servID,
		manager:    manager,
		Dispatcher: network.NewBlockingDispatcher(),
	}
}
```

And once the methods themselves are defined and call another part of the structure:

```
// RegisterProcessor overrides the RegisterProcessor methods of the Dispatcher.
// It delegates the dispatching to the serviceManager.
func (c *Context) RegisterProcessor(p network.Processor, msgType network.MessageTypeID) {
	c.manager.registerProcessor(p, msgType)
}

// RegisterProcessorFunc takes a message-type and a function that will be called
// if this message-type is received.
func (c *Context) RegisterProcessorFunc(msgType network.MessageTypeID, fn func(*network.Envelope)) {
	c.manager.registerProcessorFunc(msgType, fn)
}
```

Which one should have precedence? Can we delete one of the two references?

jeffallen:
Also: Remove an unused field

Fixes #230.

### DIFF
--- a/context.go
+++ b/context.go
@@ -15,7 +15,6 @@ type Context struct {
 	serviceID  ServiceID
 	manager    *serviceManager
 	bucketName string
-	network.Dispatcher
 }
 
 // defaultContext is the implementation of the Context interface. It is
@@ -27,7 +26,6 @@ func newContext(c *Server, o *Overlay, servID ServiceID, manager *serviceManager
 		serviceID:  servID,
 		manager:    manager,
 		bucketName: ServiceFactory.Name(servID),
-		Dispatcher: network.NewBlockingDispatcher(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -15,7 +15,6 @@ type Context struct {
 	serviceID  ServiceID
 	manager    *serviceManager
 	bucketName string
-	network.Dispatcher
```
ineiti: I'm not sure all parts of cothority still work with this change! Can you verify that, please?

jeffallen: Cothority builds and passes "make test" with this change in place.

PR #349, opened 2018-02-26
Add an expiration mechanism to the TreeNodeCache

### LINKED ISSUE #271: Expire TreeNodeCache entries
The TreeNodeCache grows without bound right now. Enrties in it should expire.

Same might be true for the roster cache, according to @ineiti , but I can't find it.

jeffallen:
Fixes #271.

### DIFF
--- a/overlay.go
+++ b/overlay.go
@@ -327,7 +327,7 @@ func (o *Overlay) TreeNodeFromToken(t *Token) (*TreeNode, error) {
 	if tn == nil {
 		return nil, errors.New("didn't find treenode")
 	}
-	// Since we found treeNode, cache it so later reuse
+	// Since we found treeNode, cache it.
 	o.cache.Set(tree, tn)
 	return tn, nil
 }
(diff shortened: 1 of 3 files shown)


overlay.go
```diff
@@ -322,7 +328,7 @@ func (o *Overlay) TreeNodeFromToken(t *Token) (*TreeNode, error) {
 		return nil, errors.New("didn't find treenode")
 	}
 	// Since we found treeNode, cache it so later reuse
```
ineiti: s/so later reuse/so we can reuse it later/

overlay.go
```diff
@@ -663,64 +669,104 @@ type pendingMsg struct {
 	MessageProxy
 }
 
-// TreeNodeCache is a cache that maps from token to treeNode. Since the mapping
+// treeNodeCache is a cache that maps from token to treeNode. Since the mapping
 // is not 1-1 (many Token can point to one TreeNode, but one token leads to one
```
ineiti: s/many Token/many Tokens/

overlay.go
```diff
@@ -663,64 +669,104 @@ type pendingMsg struct {
...
-	if !ok {
-		// no tree cached for this token :...
+	ce, ok := tnc.Entries[tok.TreeID]
+	if !ok || time.Now().After(ce.expiration) {
+		// no tree cached for this token
 		return nil
 	}
-	tn, ok := mm[tok.TreeNodeID]
+
+	tn, ok := ce.treeNodeMap[tok.TreeNodeID]
```
ineiti: Shouldn't we update the expiration time when it is accessed?

PR #350, opened 2018-02-27
Use []byte for bucketName/Load/Save

kc1212:
Use `[]byte` for `bucketName` to avoid alocation and use `[]byte` in the DB wrapper functions such as Load/Save.

This change breaks cothority, so we'd want to merge https://github.com/dedis/cothority/pull/1090 quickly after this one

### DIFF
--- a/context.go
+++ b/context.go
@@ -14,7 +14,7 @@ type Context struct {
 	server     *Server
 	serviceID  ServiceID
 	manager    *serviceManager
-	bucketName string
+	bucketName []byte
 }
 
 // defaultContext is the implementation of the Context interface. It is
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -164,10 +164,10 @@ func (c *Context) Load(key string) (interface{}, error) {
 // This function should only be used if the Load and Save functions are not sufficient.
 // Additionally, the user should not create buckets directly on the DB but always
 // call this function to create new buckets to avoid bucket name conflicts.
-func (c *Context) GetAdditionalBucket(name string) (*bolt.DB, string) {
-	fullName := c.bucketName + "_" + name
+func (c *Context) GetAdditionalBucket(name []byte) (*bolt.DB, []byte) {
```
jeffallen: I'd be OK with this staying string, if you thought it makes the API nicer.

kc1212: Since bucketName is already []byte, I thought it'll be consistent for this one to be []byte too.

jeffallen: OK, merging.

PR #353, opened 2018-02-28
Prepare for v2.

jeffallen:


ineiti:
I would like to keep the Makefile for the stable version - there is no reason to test creation of stable _from_ the stable repo. Which could happen if we PR against `v2` and travis will run.

jeffallen:
The stable/overwrite/.travis.yml file needs to be there anyway for the go_import_path, and so in it we call "make test_stable" instead of "make test". Then there is no need to maintain two copies of Makefile.

PR #354, opened 2018-02-28
Add CounterIO interface to TreeNodeInstance and Overlay

kc1212:
This PR changes the `Conn` interface where `Send() error` is now `Send (uint64, error)`, and adds the CounterIO interface to TreeNodeInstance and Overlay.

It makes https://github.com/dedis/onet/pull/265 and https://github.com/dedis/onet/pull/264 redundant as it should be an accurate way to count tx/rx. However, I did not add a CounterIO interface to Context because there wasn't an obvious way to do it and Crux only needs it on TreeNodeInstance. Further, since Crux is a long-term project and should be eventually be ported to onet.v2, I do not plan to backport this change to onet.v1.

### DIFF
--- a/overlay.go
+++ b/overlay.go
@@ -264,6 +264,7 @@ func (o *Overlay) requestTree(si *network.ServerIdentity, onetMsg *ProtocolMsg, 
 		return err
 	}
 
+	// no need to record sentLen because Overlay uses Server's CounterIO
 	_, err = o.server.Send(si, msg)
 	return err
 }
--- a/overlay.go
+++ b/overlay.go
@@ -264,7 +264,7 @@ func (o *Overlay) requestTree(si *network.ServerIdentity, onetMsg *ProtocolMsg, 
 		return err
 	}
 
-	err = o.server.Send(si, msg)
+	_, err = o.server.Send(si, msg)
 	return err
 }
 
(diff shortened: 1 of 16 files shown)


overlay.go
```diff
@@ -264,7 +264,8 @@ func (o *Overlay) requestTree(si *network.ServerIdentity, onetMsg *ProtocolMsg,
 		return err
 	}
 
-	err = o.server.Send(si, msg)
+	// no need to record sentLen because Overlay uses Server's CounterIO
+	_, err = o.server.Send(si, msg)
```
ineiti: Couldn't there be more than one overlay? Then the number of sent bytes in Overlay and in Server would be different. But if 1 Server has only 1 Overlay, then this is OK.

kc1212: I will think about this one.

kc1212: I don't think we have a straightforward solution for this problem. I could do something like `reflect.TypeOf(Envelope).Size()`, but that may not give us the right results because fields in a message may be pointers. Another solution is to change more of the API, i.e.
```
type Conn interface {
	...
	Receive() (*Envelope, uint64, error) // add uint64
	...
}

type Dispatcher interface {
	RegisterProcessor(p Processor, msgType ...MessageTypeID)
	RegisterProcessorFunc(MessageTypeID, func(*Envelope))
	Dispatch(*Envelope, uint64) error // add uint64
}

type Processor interface {
	Process(*Envelope, uint64) // add uint64
}
```

These changes are needed because router calls `Receive`, which would return the message size, then it calls `Dispatch` and we can pass the size down. Finally, `Dispatch` calls `Process`, which is implemented by `Overlay`, and we can record the Rx.

kc1212: How does this proposal sound @ineiti ? Another way is to just remove CounterIO from Overlay, because Crux doesn't need it (it only needs it on TreeNodeInstance).

ineiti: I'm still not sure my comment is even correct ;) Is there a way to have more than one overlay per server?

kc1212: Actually, I should've checked that. In `server.go:newServer`, it does `c.overlay = NewOverlay(c)`, so the answer is no.

kc1212: I don't `c.overlay` is replaced anywhere else.

PR #357, opened 2018-03-02
Improve websocket errors

ineiti:
Helping students find their errors...

### DIFF
--- a/processor_test.go
+++ b/processor_test.go
@@ -76,12 +76,20 @@ func TestServiceProcessor_ProcessClientRequest(t *testing.T) {
 
 	buf, err = protobuf.Encode(&testMsg{42})
 	log.ErrFatal(err)
-	rep, err = p.ProcessClientRequest(nil, "testMsg", buf)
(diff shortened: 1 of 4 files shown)


processor_test.go
```diff
@@ -76,12 +76,20 @@ func TestServiceProcessor_ProcessClientRequest(t *testing.T) {
...
 	log.ErrFatal(err)
-	rep, err = p.ProcessClientRequest(nil, "testMsg2", buf)
+	_, err = p.ProcessClientRequest(nil, "testMsg2", buf)
+	assert.NotNil(t, err)
+
+	// Test non-existing endpoint
+	buf, err = protobuf.Encode(&testMsg2{42})
+	log.ErrFatal(err)
+	log.OutputToBuf()
+	_, err = p.ProcessClientRequest(nil, "testMsgNotAvailable", buf)
```
jeffallen: Were you planning on adding a call to GetStdOut() and string.Contains here to check the log? It kind of looks like it, but it's not so important.

PR #360, opened 2018-03-06
more logging for startup

ineiti:
Log when a conode starts.


kc1212:
Something wrong with the coveralls, but merging anyway.

PR #362, opened 2018-03-07
Add control of the address the server is listening on

### LINKED ISSUE #176: It should be possible to control the net.Listen() address
Now: `NewTCPListener` in `onet/network/tcp.go` always binds globally.

Problems: 
- for a good firewall, an app trying to bind globally is refused
- not globally binding doesn't work when the nodes are running in a virtual machine

Proposed solution:
- `NetTCPListener` checks the address: if it's a private address, it binds to it, if it's a public address, it binds globally. Use `Address.Public()` to decide if it's public or not.

ValentinMoullet:
Add ListenAddress field to private.toml and to the CothorityConfig struct to be used to control the address we are listening on in NewTCPListener.

Not sure about if it should be tested more. Also, not sure if we should have a ListenAddress in the LoadSimulationConfig function, in simulation.go.

Fixes #176

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -17,12 +17,18 @@ import (
 )
 
 // CothorityConfig is the configuration structure of the cothority daemon.
+// - Suite: The cryptographic suite
+// - Public: The public key
+// - Private: The Private key
+// - Address: The external address of the conode, used by others to connect to this one
+// - ListenAddress: The address this conode is listening on
+// - Description: The description
 type CothorityConfig struct {
 	Suite         string
 	Public        string
 	Private       string
 	Address       network.Address
-	ListenAddress network.Address
+	ListenAddress string
 	Description   string
 }
 
--- a/app/config.go
+++ b/app/config.go
@@ -18,11 +18,12 @@ import (
 
 // CothorityConfig is the configuration structure of the cothority daemon.
 type CothorityConfig struct {
-	Suite       string
-	Public      string
-	Private     string
-	Address     network.Address
-	Description string
+	Suite         string
+	Public        string
+	Private       string
+	Address       network.Address
+	ListenAddress network.Address
+	Description   string
 }
 
 // Save will save this CothorityConfig to the given file name. It
@@ -73,7 +74,7 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
 	si := network.NewServerIdentity(point, hc.Address)
 	si.SetPrivate(private)
 	si.Description = hc.Description
-	server := onet.NewServerTCP(si, suite)
+	server := onet.NewServerTCP(si, suite, hc.ListenAddress)
 	return hc, server, nil
 }
 
--- a/app/config.go
+++ b/app/config.go
@@ -80,7 +80,7 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
 	si := network.NewServerIdentity(point, hc.Address)
 	si.SetPrivate(private)
 	si.Description = hc.Description
-	server := onet.NewServerTCP(si, suite, hc.ListenAddress)
+	server := onet.NewServerTCPWithListenAddr(si, suite, hc.ListenAddress)
 	return hc, server, nil
 }
 
--- a/app/config.go
+++ b/app/config.go
@@ -80,7 +80,13 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
 	si := network.NewServerIdentity(point, hc.Address)
 	si.SetPrivate(private)
 	si.Description = hc.Description
-	server := onet.NewServerTCPWithListenAddr(si, suite, hc.ListenAddress)
+	server := onet.NewServerTCP(si, suite)
+	if hc.ListenAddress != "" {
+		err := server.SetListenAddress(hc.ListenAddress)
+		if err != nil {
+			return nil, nil, fmt.Errorf("setting listen address: %v", err)
+		}
+	}
 	return hc, server, nil
 }
 
--- a/app/config.go
+++ b/app/config.go
@@ -80,13 +80,8 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
 	si := network.NewServerIdentity(point, hc.Address)
 	si.SetPrivate(private)
 	si.Description = hc.Description
-	server := onet.NewServerTCP(si, suite)
-	if hc.ListenAddress != "" {
-		err := server.SetListenAddress(hc.ListenAddress)
-		if err != nil {
-			return nil, nil, fmt.Errorf("setting listen address: %v", err)
-		}
-	}
+	// Same as `NewServerTCP` if `hc.ListenAddress` is empty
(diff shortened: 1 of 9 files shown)


kc1212:
This PR seems to change some public functions. Can you please check if cothority uses these functions and update them if necessary?

ValentinMoullet:
@kc1212 I checked and cothority doesn't use any of those functions directly + the tests are still passing.

app/config.go
```diff
@@ -18,11 +18,12 @@ import (
...
-	Suite       string
-	Public      string
-	Private     string
-	Address     network.Address
-	Description string
+	Suite         string
+	Public        string
+	Private       string
+	Address       network.Address
+	ListenAddress network.Address
```
jeffallen: I'm surprised to see a onet/network.Address here. I was expecting this to either be a string, or a net.IP (i.e. already parsed, perhaps via some magic in the toml parser).

Now, it is true that onet/network.Address is in fact just a string, so it's no big deal. But having "tcp://$ip:$port" here, as the documentation implies, doesn't make sense; the syntax of ListenAddress is one single IP address (v4 or v6).

network/tcp.go
```diff
@@ -265,7 +265,7 @@ type TCPListener struct {
 // the binding.
 // A subsequent call to Address() gives the actual listening
 // address which is different if you gave it a ":0"-address.
-func NewTCPListener(addr Address, s Suite) (*TCPListener, error) {
+func NewTCPListener(addr Address, s Suite, listenAddr Address) (*TCPListener, error) {
```
jeffallen: There is a magical default for listenAddr, in this function. It must be documented.

network/tcp_test.go
```diff
@@ -363,10 +363,48 @@ func TestTCPConnWithListener(t *testing.T) {
 	}
 }
 
-// will create a TCPListener & open a golang net.TCPConn to it
+// will create a TCPListener with a specific address to listen on &
+// open a golang net.TCPConn to it
+func TestTCPListenerWithListenAddr(t *testing.T) {
+	addr := NewAddress(PlainTCP, "127.0.0.1:5678")
+	listenAddr := NewAddress(PlainTCP, "127.0.0.1:5670")
```
jeffallen: Whoa, no no. ListenAddr is not for changing the port number, it is only for changing the IP address in the second argument to net.Listen in network/tcp.go, line 285.

Sorry to deliver this big correction only now...

network/tcp_test.go
```diff
@@ -363,10 +363,48 @@ func TestTCPConnWithListener(t *testing.T) {
 	}
 }
 
-// will create a TCPListener & open a golang net.TCPConn to it
+// will create a TCPListener with a specific address to listen on &
+// open a golang net.TCPConn to it
+func TestTCPListenerWithListenAddr(t *testing.T) {
+	addr := NewAddress(PlainTCP, "127.0.0.1:5678")
```
jeffallen: A test that listens on a fixed port will give false negatives if that port is already in use. Fragile tesdt = crazy mad Jeff. :)

Instead, listen on port 0, so that the kernel gets to choose. Use "addr := NewAddress(PlainTCP, "1.2.3.4:0") and listenAddr := net.ParseIP("127.0.0.1")". Then connect to it via 127.0.0.1 (use svr.Address() to find out the address that the kernel assigned). If you succeed, it means the listenAddr worked (otherwise, it would be trying and failing to listen on 1.2.3.4).

network/tcp_test.go
ineiti: I'm not sure that we don't want to change the listening port, too. In some docker setups you might want to have another 'public' port-#, compared to the listening port-#. So I suggest to keep that functionality. We can add some magic and say that:
1. if there is no `listenAddr`, take the `ServerIdentity.Address`
2. if there is an IP in `listenAddr`, take the port from `ServerIdentity.Address`
3. if there is an IP:Port in `listenAddr`, take this

@ValentinMoullet didn't want to mix IP from `listenAddr` and port from `ServerIdentity.Address`, which is also a valid concern.

network/tcp_test.go
```diff
@@ -363,10 +363,48 @@ func TestTCPConnWithListener(t *testing.T) {
...
+	listenAddr := NewAddress(PlainTCP, "127.0.0.1:5670")
+	ln, err := NewTCPListener(addr, tSuite, listenAddr)
+	if err != nil {
```
ineiti: `require.Nil(t, err)`

app/config.go
```diff
@@ -18,11 +18,12 @@ import (
 
 // CothorityConfig is the configuration structure of the cothority daemon.
 type CothorityConfig struct {
```
ineiti: Add comments to each field

network/tcp_test.go
ValentinMoullet: Alright, makes sense. In this case, should we also change the other tests in this file listening on a fixed port (e.g. TestTCPListener, TestTCPConnWithListener, TestTCPConn)? If yes, should I do it for this PR?

network/tcp_test.go
ValentinMoullet: Exactly. Why would having ListenAddr to change the port be bad @jeffallen ?

For your suggestion @ineiti , I agree that this would make sense, and in the end I'm ok with "mixing" them in the case there is only an IP. All this magic will of course need to be commented.

network/tcp_test.go
```diff
@@ -373,13 +373,71 @@ func TestTCPConnWithListener(t *testing.T) {
 	}
 }
 
-// will create a TCPListener & open a golang net.TCPConn to it
+// will create a TCPListener with a specific address to listen on &
```
ineiti: `s/&/and/`

network/tcp_test.go
```diff
@@ -373,13 +373,71 @@ func TestTCPConnWithListener(t *testing.T) {
...
+	addr := NewAddress(PlainTCP, "1.2.3.4:0")
+	listenAddr := ":0"
+	_, err := NewTCPListenerWithListenAddr(addr, tSuite, listenAddr)
+	require.NotNil(t, err, "Should fail: cannot only have a port as a listen address")
+
+	// Testing different working configurations.
+	addrs := [2]Address{NewAddress(PlainTCP, "1.2.3.4:0"), NewAddress(PlainTCP, "1.2.3.4:0")}
+	listenAddrs := [2]string{"127.0.0.1", "127.0.0.1:0"}
+	expectedHosts := [2]string{"127.0.0.1", "127.0.0.1"}
+	for i := 0; i < len(addrs); i++ {
```
ineiti: Something like this is more go-like:
```
testVector := struct{addr Address; listen string; expected string}{{NewAddress(...), "127.0.0.1", "127.0.0.1"}, {}}
for _, tv := range testVector{
  ln, err := NewTCPListenerWithListenAddr(tv.addr, tSuite, tv.listen)
...
```

see also: https://github.com/dedis/onet/blob/cd17dbbe5846e58a8194c1bc63ebed142cbed454/network/address_test.go#L51-L85

ineiti:
@jeffallen - is it OK for you now?

PR #364, opened 2018-03-09
First try at removing some overlay memory leaks.

jeffallen:
@kc1212 and @ineiti found that trees and rosters were very badly set up:
- every time `NewRosterWithRoot` is called, the roster gets a new, random ID
- every time `GenerateNaryTreeWithRoot` is called, even for the same tree, it gets a random ID

-> every call for a new skipchain-block will create a new roster/tree - and they are not cleaned up!

Now we changed it so that rosters depend on the order of the ServerIdentities, and trees depend on the roster and the order of the nodes in the tree. But no random IDs anymore.

### DIFF
--- a/local.go
+++ b/local.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"testing"
 	"time"
 
 	"github.com/dedis/kyber"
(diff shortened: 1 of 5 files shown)


local.go
```diff
@@ -171,14 +173,28 @@ func (l *LocalTest) panicClosed() {
...
+// CloseAll closes all the servers.
 func (l *LocalTest) CloseAll() {
 	log.Lvl3("Stopping all")
 	// If the debug-level is 0, we copy all errors to a buffer that
 	// will be discarded at the end.
 	if log.DebugVisible() == 0 {
 		log.OutputToBuf()
 	}
+
+	if l.T != nil {
```
jeffallen: I did it this way to make it possible for protocol tests to "opt-in" to leak checking for the moment. That way we don't have too much stuff crashing all over the place while we experiment with this.

PR #368, opened 2018-03-16
ignoring EOF in new connections

ineiti:
If a load balancer connects to our conode-to-conode communication, it will close the connection without sending anything. So ignore EOF errors in the server identity verification.

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/dedis/kyber/util/encoding"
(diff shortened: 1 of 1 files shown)


kc1212:
Why does a load balancer have this behaviour?

network/router.go
```diff
@@ -101,7 +102,12 @@ func (r *Router) Start() {
 	err := r.host.Listen(func(c Conn) {
 		dst, err := r.receiveServerIdentity(c)
 		if err != nil {
-			log.Error("receive server identity failed:", err)
+			if strings.Contains(err.Error(), "EOF") {
+				// This is probably just a stray connection - ignore it.
+				return
```
kc1212: Do we want to close the connection just to make sure?

ineiti: Something like that? Or will it print an error message if the other end closed the connection and we close it, too?

PR #373, opened 2018-03-16
Fatal when calling GlobalProtocolRegister with server(s) already started

ValentinMoullet:
By using a flag to `protocolStorage`, we can now check that no server has already been started while calling `GlobalProtocolRegister`, and if this is the case we `log.Fatal`.

Didn't add a test for this change. Should we? If yes, how to test it cleanly since we use `log.Fatal`?

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -55,10 +55,16 @@ var protocols = newProtocolStorage()
 
 // protocolStorage holds all protocols either globally or per-Server.
 type protocolStorage struct {
+	// Lock used because of the 'serverStarted' flag: it can be changed from a
+	// call to 'Server.Start' and is checked when calling
+	// 'GlobalProtocolRegister'.
+	sync.Mutex
 	// Instantiators maps the name of the protocols to the `NewProtocol`-
 	// methods.
 	instantiators map[string]NewProtocol
-	sync.Mutex
+	// Flag indicating if a server has already started; here to avoid calls
+	// to 'GlobalProtocolRegister' when a server has already started.
+	serverStarted bool
 }
 
 // newProtocolStorage returns an initialized ProtocolStorage-struct.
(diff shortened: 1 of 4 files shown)


protocol.go
```diff
@@ -109,9 +116,28 @@ func ProtocolNameToID(name string) ProtocolID {
 // All registered protocols will be copied to every instantiated Server. If a
 // protocol is tied to a service, use `Server.ProtocolRegisterName`
 func GlobalProtocolRegister(name string, protocol NewProtocol) (ProtocolID, error) {
+	protocols.Lock()
+	defer protocols.Unlock()
+	if protocols.serverStarted {
+		log.Lvl4("Cannot call 'GlobalProtocolRegister' when a server has already started.")
```
ineiti: Shouldn't this do `return nil, errors.New()`?

ValentinMoullet: In fact I wanted to put `log.Fatal` at first (I don't know why I ended up putting `log.Lvl4`), to fail directly. The main reason was because this function is currently used a lot in Cothority (and Onet), and it (almost) never checks the returned error, so returning the error would currently not do anything. Also, one of @jeffallen comment on the issue was: "Goal: give early, clear crash/error so that students don't fall into this trap.".

I could also return the error as you suggest, but then, before that, I'll have to go over all the calls to this function in Cothority and Onet and add the code to check the returned error, and fail/return the error further.

What do you think makes more sense in our case?

ineiti: that's horrible! I feel very bad ;) Then do a `panic` here, so you can catch it in the tests and look if it behaves as needed: https://blog.golang.org/defer-panic-and-recover

ineiti:
Great work - if I interpret your commit-history correctly, you used `git merge` repeatedly. We usually do `git fetch; git rebase origin/master; git push --force` for a PR - even if it rewrites history, but as it's only for that branch, we do accept it.

This makes the commit-history cleaner.

ineiti:
You should definitively use `git rebase`, else we have too many non-relevant changes in here.

ValentinMoullet:
I'm using `git rebase` generally, but this time I couldn't push anymore, so I had to `git pull` to "fix" this problem, I didn't know about `git push --force` which would have solved my problem without adding all those non-relevant changes I think. Thank you for the tip!

ValentinMoullet:
All tests passed in Cothority as well with these new changes.

PR #377, opened 2018-03-21
Improve testing

ineiti:
- `TreeNodeInstance.Info()`: print protocol name
- `LocalTest`: allow list of all `TreeNodeInstances` in tests

### DIFF
--- a/local.go
+++ b/local.go
@@ -260,11 +260,11 @@ func (l *LocalTest) NewTreeNodeInstance(tn *TreeNode, protName string) (*TreeNod
 	return node, nil
 }
 
-// getNodes returns all Nodes that belong to a treeNode
-func (l *LocalTest) getNodes(tn *TreeNode) []*TreeNodeInstance {
+// GetTreeNodeInstances returns all TreeNodeInstances that belong to a server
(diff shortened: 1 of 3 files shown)


local.go
```diff
@@ -260,11 +260,11 @@ func (l *LocalTest) NewTreeNodeInstance(tn *TreeNode, protName string) (*TreeNod
 	return node, nil
 }
 
-// getNodes returns all Nodes that belong to a treeNode
-func (l *LocalTest) getNodes(tn *TreeNode) []*TreeNodeInstance {
+// GetTreeNodeInstances returns all TreeNodeInstances that belong to a server
+func (l *LocalTest) GetTreeNodeInstances(id network.ServerIdentityID) []*TreeNodeInstance {
```
kc1212: Why does this need to be public?

ineiti: so that tests can ask what treenodeinstances are still alive.

In a second step I would like to add something like the following to `LocalTest.CloseAll`:

```
	for _, s := range o.local.Servers {
		for _, tn := range o.local.GetTreeNodeInstances(s.ServerIdentity.ID) {
			log.Lvl1("Still active: ", tn.Info())
		}
	}
```

PR #381, opened 2018-03-21
Simulations using TCP by default, with flag for TLS

ValentinMoullet:
- Created `NewTCPAddress` function in `network/tcp.go`.
- Changed a maximum of TCP `Address` creation with this new function.
- By default, using `PlainTCP` addresses in simulations, instead of `TLS`.
- Added a `TLS` flag `SimulationConfig`, `SimulationConfigFile` and `SimulationBFTree` structures to let the user decide if `TLS` should be used instead of the default `PlainTCP`.

It seems to me that it is necessary to add the `TLS` flag to those 3 structures, maybe it's good if you check that this is indeed the case.

Also, the current tests in `simulation_test.go` can all be run using `PlainTCP`, right?

### DIFF
--- a/app/config_test.go
+++ b/app/config_test.go
@@ -39,7 +39,7 @@ func TestReadGroupDescToml(t *testing.T) {
 		t.Fatal("Should have 2 ServerIdentities")
 	}
 	nikkoAddr := group.Roster.List[0].Address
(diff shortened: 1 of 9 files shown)


ineiti:
Can you do a `make test` in the cothority-repository using that code? I'd hate to introduce regressions here...

ValentinMoullet:
All tests passed in Cothority as well with these new changes.

PR #400, opened 2018-04-10
Add Go module info to status.

jeffallen:
Extract the Go module information from the binary in order to make it available in the status reply. This Go module information will be embedded in binaries which are built with vgo. In case that parsing the information fails, or the module information is not present, then the status reply is unchanged.


kc1212:
I don't really understand what the `go.mod` and `go.modverify` are for, or how they're updated.

jeffallen:
go.mod and go.modverify are new and experimental. go.mod is like Gopkg.lock file. go.modverify holds  hashes of the source as it was fetched, so that others who try to build the same code can find out if they are building from the same source. 

It is not yet clear if onet should really have these, or only cothority.

PR #401, opened 2018-04-11
Use sha256(pk) as the database name

### LINKED ISSUE #399: Database file names should be sha256 of the public key
Currently the database names are simply the public keys of the service. However, when using bn265 as the identity, where the public key is very long, the database cannot be created because it's hitting the filesystem filename length limit. The solution would be to use sha256(pk) as the database file name.

This is a non-backward compatible change because existing systems would not be able to open old database files after the upgrade. We need mechanism to upgrade databases. One proposoal si to have the service try to open the database with the new name, i.e. sha256(pk), if it fails, then it will try to open it using the old name and make a copy of the database under the new name.

It is highly unlikely that there would be a collision, i.e. sha256(pk) == pk, so we do not need to introduce a version field in the database yet.

kc1212:
We do this because, depending on the suites, public keys may be very
long, too long for filesystems. This commit uses the sha256 of the
public key as the database file name and automatically converts old
filenames to the new one.

This feature is a bit difficult to test by unit tests. We have integration
test in cothority/conode. 

Fixes #399

### DIFF
--- a/service.go
+++ b/service.go
@@ -289,7 +289,7 @@ func (s *serviceManager) dbFileName() string {
 	return path.Join(s.dbPath, fmt.Sprintf("%x.db", h.Sum(nil)))
 }
 
-// updateDbFileName checks if the old database file name exists, it does, it
+// updateDbFileName checks if the old database file name exists, if it does, it
 // will rename it to the new file name.
 func (s *serviceManager) updateDbFileName() {
 	if _, err := os.Stat(s.dbFileNameOld()); err == nil {
--- a/service.go
+++ b/service.go
@@ -1,13 +1,13 @@
 package onet
 
 import (
+	"crypto/sha256"
 	"errors"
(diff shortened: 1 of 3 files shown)


service.go
```diff
@@ -275,11 +277,30 @@ func createBucketForService(db *bolt.DB, bucketName string) error {
...
 }
 
+func (s *serviceManager) dbFileName() string {
+	pub, _ := s.server.ServerIdentity.Public.MarshalBinary()
+	h := sha256.New()
+	h.Write(pub)
+	return path.Join(s.dbPath, fmt.Sprintf("%x.db", h.Sum(nil)))
+}
+
+// updateDbFileName checks if the old database file name exists, it does, it
```
ineiti: s/it does/if it does/

kc1212:
Fixed the typo

PR #402, opened 2018-04-16
Change to the real goversion, now that v1.0.2 is released.

jeffallen:


kc1212:
Ops, the commit message is wrong, it should be v1.2.0. Did not realise it earlier, probably not a big deal.

PR #406, opened 2018-04-25
Router should not open a connection when sending message to itself.

ValentinMoullet:
`Router.Send` is now checking if we are sending a message to ourselves, and if this is the case, it will directly dispatch the message without opening a connection with ourselves.

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -166,6 +166,27 @@ func (r *Router) Send(e *ServerIdentity, msg Message) (uint64, error) {
 		return 0, errors.New("Can't send nil-packet")
 	}
 
+	// If sending to ourself, directly dispatch it
+	if e.Address == r.address {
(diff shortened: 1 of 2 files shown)


network/router.go
```diff
@@ -166,6 +166,27 @@ func (r *Router) Send(e *ServerIdentity, msg Message) (uint64, error) {
 		return 0, errors.New("Can't send nil-packet")
 	}
 
+	// If sending to ourself, directly dispatch it
+	if e.Address == r.address {
```
ValentinMoullet: One question I have: is this check ok? Or should we compare the `NetworkAddress()`?

ineiti: Both addresses are strings, so this is OK. We could think about using the IP addresses, but they might not resolve.

PR #417, opened 2018-05-03
adding mutex to Localhost

ineiti:
Because simul/build.go has a global holding the test and uses it in different
simulations, we need to hold a mutex, at least for localhost.


kc1212:
Does this allow us to run tests without `-p=1`?

PR #424, opened 2018-05-08
Comparing ServerIdentityID when checking if sending package to ourself.

### LINKED ISSUE #423: Compare serverID's via x.ID == y.ID
In this check for "are we trying to connect to ourselves?":

https://github.com/dedis/onet/blob/b2cbe3e6d3067f9275b52fbca979efae49e5d2d0/network/router.go#L170

the check can fail to trigger if the address is written as both an IP and also as a FQDN. Instead, we should be comparing the ID field.

ValentinMoullet:
Comparing addresses can lead to issues due to canonicalisation. Comparing `ServerIdentity`'s ID to know if we are sending a message to ourself is thus better.

Fixes #423

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -168,7 +168,7 @@ func (r *Router) Send(e *ServerIdentity, msg Message) (uint64, error) {
 	}
 
 	// If sending to ourself, directly dispatch it
-	if e.Address == r.address {
+	if e.ID == r.ServerIdentity.ID {
 		log.Lvlf4("Sending to ourself (%s) msg: %+v", e, msg)
 		packet := &Envelope{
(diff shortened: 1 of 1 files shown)


network/router.go
```diff
@@ -168,7 +168,7 @@ func (r *Router) Send(e *ServerIdentity, msg Message) (uint64, error) {
 	}
 
 	// If sending to ourself, directly dispatch it
-	if e.Address == r.address {
+	if e.ID == r.ServerIdentity.ID {
```
kc1212: we should use `e.ID.Equal` to do a deep comparison.

PR #427, opened 2018-05-16
Refactored log-library + added possibility to create a file logger and a syslog logger

ValentinMoullet:
- Extended/modified the `Listener` interface to a new one, `Logger`, as well as the possibility to unregister loggers.
- Created a structure called `LoggerInfo` to store loggers information and that embeds `Logger`; the goal is that all loggers should use this structure and set the fields + `Logger` as they wish.
- Created `NewFileLogger` which creates and registers a logger that writes into a given file.
- Created `NewSyslogLogger` which creates and registers a logger that writes into syslog.
- Refactored `lvl.go` so that we first initialise and register a standard logger writing to `stdOut` and `stdErr`; then `lvl` now goes through all registered loggers (including the standard one) and creates the string to log following the same rules, each time depending on the info in `LoggerInfo`, such as `debugLvl`, `showTime` and `useColors`.

Not sure about:
- The padding. For the standard logger, we were logging string like this one: "I : ( [LOTS OF SPACE TO PAD] log.TestLvl:   0) - TestLvl". Do we want this for all the loggers? Should this be another field in `LoggerInfo`?
- `print` in `ui.go`. Right now, it's only outputting using the standard logger, but do we want that, or do we want `print` to use all registered loggers?


log/lvl.go
```diff
@@ -77,108 +64,124 @@ var debugMut sync.RWMutex
...
+
+		if l.useColors {
+			ct.ResetColor()
+		}
 	}
 }
 
-func fg(c ct.Color, bright bool) {
-	if useColors {
+func fg(l *LoggerInfo, c ct.Color, bright bool) {
```
ineiti: This will now be called by every Logger, and the corresponding `ct.ResetColor` also. Better is to put these calls into `stdLogger.Log`, then only this logger will set/reset the colors.

log/lvl.go
```diff
@@ -77,108 +64,124 @@ var debugMut sync.RWMutex
...
 
 func init() {
+	initStdLogger()
 	ParseEnv()
 }
 
-func lvl(lvl, skip int, args ...interface{}) {
-	debugMut.Lock()
-	defer debugMut.Unlock()
+type stdLogger struct{}
```
ineiti: Now we have three types of loggers, so I vote for putting all three of them into a `loggers.go` file separately.

ineiti:
And yes, I think it makes sense to have a `DEBUG_PADDING` variable in addition to `DEBUG_LVL` and `DEBUG_COLOR`.

log/lvl.go
ValentinMoullet: The problem is that I don't have access to the `LoggerInfo` (and thus `useColors`) from the `stdLogger.Log`. Thus I cannot call `fg` from here. If we want to only have those colors for the `stdLogger`, one way would be to check that the logger we are iterating on is the first one (index 0) when we do those color changes, we already rely on this fact anyway. Or maybe there you can think of another solution?

ineiti: True - I supposed that the `LoggerInfo` _are_ the loggers, which is not the case. I can think of two ways:

1. Create a `Logger`-interface that as `SetDebugVisible`, `GetDebugVisible`... and  then it would be `NewFileLogger(name string)(int, error)`, `Logger` would be a simple `io.WriteCloser`, and you would have a `Logger.SetColor` that would be empty for most loggers.

2. Change the signature of `Logger.Log` to `Log(msg string, color ct.Color)`, because you don't need the `lvl`, but only the color.

I have a small preference for 1., but 2. seems to be OK to me, too.

ValentinMoullet: For 1., I'm not sure I understand the interface/struct logic you describe. What would be wrapping what? Where do I store the info about each logger (`debugLvl`, `showTime`, ...)? Do we have 2 interfaces, one being a `io.WriteCloser` and another one having `SetDebugVisible `, `SetUseColors`, etc... ?

For 2., how do I differentiate between cases where I have to write in stdOut or stdErr? It seems to me that I still need the `lvl` (in the case of the `stdLogger` at least). And if we do that, in order to check which `color` I should give, I'd still have to check if this is the `stdLogger` or not (checking the logger's index in `loggers`). Or maybe I misunderstood a part?

I have a 3rd alternative: in `NewFileLogger` and `NewSyslogLogger`, we explicitly put the `useColors` to `false`. In general, if you want to force a certain value for any `Logger`, you set it in the function used to create this `Logger`.

ineiti: 1. each logger would store the info itself. Instead of reading out the field, `log` would have to call `Logger.GetDebugVisible`. The `Logger` interface would also implement the `io.WriteCloser`, so everybody implementing `Logger` would have to add `(Set|Get)Debug(Visible|Color|Time)`, `SetColor`, `Close` and `Write`.

2. the color would be calculated in `log` and then passed to the `Logger`. All non-std loggers would simply ignore that field. But you're right, it would be better to pass the `lvl` instead of `color`, then the std-logger could set the colors appropriately and the other loggers could do something if it is a `Warn`, `Error` or `Fatal`.

3. I don't like the fact that `log` would still be handling the colors, although they will not be used for the non-std loggers. 

My new favourite is 2. - I think it would be the least invasive for what you have now and cleanly separate the colors from the non-std loggers.

ValentinMoullet: Ok, thanks for the clarifications. About 2., how do we know if we want to use colors inside `stdLogger.Log`? We don't have access to `useColors`. So I can pass the `lvl` and compute the color it would become if we want to use it, but we cannot know if we want to use it. What I could do is indeed pass the color inside `Log`, but I should also have `lvl` as mentioned before, for the differentiation between `stdOut` and `stdErr`, and maybe for other purposes in future loggers as you said.
Another solution is to put `useColors` as an argument for `Log`, and compute the color inside `stdLogger.Log`.
What do you think?

ineiti: Or you just pass the whole `LoggerInfo` to log - that's perhaps a little bit easier than 1.

ValentinMoullet: Indeed. I've done this one, so that we can use all those information as we want in `Log`.

log/loggers.go
```diff
@@ -0,0 +1,193 @@
...
+
+// LoggerInfo is a structure that should be used when creating a logger.
+// It contains parameters about how to log (with time, colors, ...) and
+// embeds the Logger interface, which should define how the logger should log.
+type LoggerInfo struct {
+	// These are information-debugging levels that can be turned on or off.
+	// Every logging greater than 'debugLvl' will be discarded . So you can
+	// Log at different levels and easily turn on or off the amount of logging
+	// generated by adjusting the 'debugLvl' variable.
+	debugLvl int
```
jeffallen: It is hard to see how a caller outside of onet would subscribe to "only messages less than or equal to 3", since debugLvl is not exported. Perhaps I am missing something about how this is designed to be used?

log/loggers.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (fl *fileLogger) Close() {
+	fl.file.Close()
+}
+
+// NewFileLogger creates and registers a logger that writes into the file with
+// the given path and is using the given LoggerInfo (without the Logger).
+// It returns the key assigned to the logger.
+func NewFileLogger(path string, lInfo *LoggerInfo) (int, error) {
```
ineiti: `LoggerInfo` is too prevalent here - it's once in `New.*` and in `Log`. Also, calling `RegisterLogger` in `New*` does more than one would expect from a `New.*` method in go.

Trying to clean up (hopefully the last time, sorry):
`LoggerInfo` does _not_ contain `Logger`. But `fileLogger`, `stdLogger` and `syslogLogger` contain `*LoggerInfo`.
All `New*` take a `*LoggerInfo` as first argument and any specific arguments necessary, they return a `Logger` interface and an error. We usually take the common argument first, and then the specific arguments. This makes it easier to remember the method signature.
`Logger.Log` does only take `lvl, msg` and uses the stored `LoggerInfo`.
`RegisterLogger` takes a `Logger` and returns an `int` for `UnregisterLogger`.

ValentinMoullet: Alright! So now, in `lvl`, I will iterate over a list of `Logger`, and not `*LoggerInfo` anymore. This means that I don't have access to things like `debugLvl`, `showTime`, etc... which is annoying since we wanted to do the generic logic of the loggers in there (e.g. checking the debug level), and the specific or more I/O related things in their `Log` method. Or am I missing something and we can easily access the `*LoggerInfo` from the `Logger`?
One solution I see is to make `Logger` implement a `GetLoggerInfo()` method returning a `*LoggerInfo`, and make all loggers implement it and just returning the `*LoggerInfo` they have.
Or maybe there is another way in Go that I don't know?

PR #429, opened 2018-05-31
unbreaking server.Start behaviour

ineiti:
server.Start is again blocking, so as not to break existing software.
There are two new methods:
- server.StartReturn calls server.Start in a goroutine and waits for all ports to be listening
- server.WaitStartup loops until the server is completely started up

### DIFF
--- a/server.go
+++ b/server.go
@@ -39,7 +39,7 @@ type Server struct {
 	// when this node has been started
 	started time.Time
 	// once everything's up and running
-	startedChannel chan bool
+	closeitChannel chan bool
 	IsStarted      bool
 
 	suite network.Suite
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -208,9 +214,35 @@ func (c *Server) Start() {
...
+	c.Lock()
 	c.IsStarted = true
+	c.Unlock()
+	// Wait for closing of the channel
+	<-c.closeitChannel
+}
+
+// StartReturn starts the services and returns once everything
+// is up and running.
+func (c *Server) StartReturn() {
```
kc1212: `StartReturn` is a bit confusing, how about `StartInBackground`?

PR #431, opened 2018-06-05
Mininet cleanup

ineiti:
Some things that went wrong when updating onet over the last weeks.
This also adds a way to ignore TCP-connections in simulations.


network/router.go
```diff
@@ -460,7 +460,9 @@ func (r *Router) receiveServerIdentity(c Conn) (*ServerIdentity, error) {
 			log.Lvl4(r.address, "Public key from CommonName and ServerIdentity match:", pub)
 		} else {
 			// We get here for TCPConn && !tls.Conn. Make them wish they were using TLS...
-			if !r.UnauthOk {
+			// Don't show error in simulations.
+			if !r.UnauthOk &&
```
jeffallen: Please arrange that UnauthOk is set to true by the simulation system. Add "c.Router.UnauthOk = true" in runsimul.go, right before c.Start()

Disabling security warnings via substring matches is not ok. :)

ineiti: OK - now I understand ;) sounds good...

jeffallen:
Forgot to "git add vartime.toml" maybe?

```
F 18/06/06 09:16:32.953111602: (                    platform.ReadRunFile: 127) - Couldn't open file <nil> open vartime.toml: no such file or directory
```

jeffallen:
Also please fix error message:
```
diff --git a/simul/platform/platform.go b/simul/platform/platform.go
index 8e5812c..5ad1a97 100644
--- a/simul/platform/platform.go
+++ b/simul/platform/platform.go
@@ -124,7 +124,7 @@ func ReadRunFile(p Platform, filename string) []*RunConfig {
 		}
 	}()
 	if err != nil {
-		log.Fatal("Couldn't open file", file, err)
+		log.Fatal("Couldn't open file", filename, err)
 	}
 
 	// Decoding of the first part of the run config file
```

PR #437, opened 2018-06-07
Allow different levels of CloseAll checks

ineiti:
This PR allows for three levels of checks:
- 0: no leaky test at all
- 1: only log.AfterTest
- 2: like 1 and test for leaky processors or protocolInstances (default)

### DIFF
--- a/local.go
+++ b/local.go
@@ -18,6 +18,20 @@ import (
 	"github.com/dedis/onet/network"
 )
 
+// LeakyTestCheck represents an enum to indicate how deep CloseAll needs to
+// check the tests.
+type LeakyTestCheck int
+
+const (
+	// CheckNone will make CloseAll not check anything.
+	CheckNone LeakyTestCheck = iota + 1
+	// CheckGoroutines will only check for leaking goroutines.
+	CheckGoroutines
+	// CheckAll will also check for leaking Overlay.Processors and
+	// ProtocolInstances.
+	CheckAll
+)
+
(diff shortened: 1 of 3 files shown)


local.go
```diff
@@ -18,6 +18,14 @@ import (
 	"github.com/dedis/onet/network"
 )
 
+// LeakyTest is false by default, but can be set to true in `TestMain`
+// to have all tests undergo more stringent testing. This only influences
+// LocalTest.CloseAll
+//  0 - no leaky test at all
```
jeffallen: Please use a typesafe version of this, i.e. `type LeakTestLevel int` and `const None iota LeakTestLevel` etc.

ineiti: Also removed the global variable, to force people to think how bad their code is if they have to write `local.CheckGoroutines` all over the place ;)

local.go
```diff
@@ -31,7 +45,7 @@ type LocalTest struct {
 	// All single nodes
 	Nodes []*TreeNodeInstance
 	// Don't test for leaky goroutines
```
jeffallen: Can you fix this comment to be: `// How carefully to check for leaking resources at the end of the test.`

ineiti:
Convinced myself that for `onet.v2` we should only have checks of go-routines in `CloseAll`, because this really helps. But not full check, as this might break things depending on `onet.v2`. And we don't want to break things depending on `onet.v2` ;)

PR #438, opened 2018-06-08
merged from latest master branch

ineiti:
Merging in things from master using `make create_stable`.


kc1212:
Does this need an equivalent one for cothority?

ineiti:
As `cothority.v2` follows `onet.v2`, it should not be needed. But there are quite some skipchain-changes - I'll do it on Monday. Or better yet, I'll show you how to do it ;)

PR #451, opened 2018-07-17
Adding a TestClose interface

ineiti:
If a service implements a `TestClose` method, it will be called in `LocalTest.CloseAll()`.

### DIFF
--- a/local.go
+++ b/local.go
@@ -32,6 +32,13 @@ const (
 	CheckAll
 )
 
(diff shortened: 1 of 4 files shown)


jeffallen:
I showed @pablo-lo why/how to do this. So it is likely repeated in his work as well, but I'll check this in here now, since it's ready.

ineiti:
I did not find any PR waiting to be merged.

PR #455, opened 2018-07-18
format the args before giving them to panic

adderbyte:
Make log.Panic  format the args before giving them to panic( ).


ineiti:
Thanks for the PR. The travis-tests should pass! Please have a look why they fail and try to fix it.

log/ui.go
```diff
@@ -37,7 +162,9 @@ func Error(args ...interface{}) {
 // Panic prints out the panic message and panics
 func Panic(args ...interface{}) {
 	lvlUI(lvlPanic, args...)
-	panic(args)
+	s := fmt.Sprint(args...)
+	std.Output(2, s)
```
jeffallen: What is this line for? I was not expecting it.

log/ui.go
```diff
@@ -14,6 +39,106 @@ func lvlUI(l int, args ...interface{}) {
 	}
 }
 
+// New creates a new Logger
+func New(out io.Writer, prefix string, flag int) *LoggerPanic {
+	return &LoggerPanic{out: out, prefix: prefix, flag: flag}
```
jeffallen: What is all of this about?

This task is about making it so that when you call something like log.Panic("The number is %d", 3) you get a panic message of "The number is 3", and not a useless message like you get today.

We do not need/want a new kind of logger, we want the current log.Panic to format the string.

jeffallen:
As a tutorial, I showed @adderbyte how to do this "my way", so closing this PR.

PR #457, opened 2018-07-25
Non-Integer parameters of simulation not in resulting csv-file #446

### LINKED ISSUE #446: Non-Integer parameters of simulation not in resulting csv-file
In https://github.com/dedis/onet/blob/master/simul/monitor/stats.go#L339, parameters of the simulation are only added to the resulting csv file when they can be parsed into an Integer.

Since other types might be used as well to parameterize a simulation, they should also be parsed.

adderbyte:
parameters of the simulation are only added to the resulting csv
 file when they can be parsed into an Integer.

Since other types might be used as well to parameterize a simulation,
 they should also be parsed.

We would like to change the parameters of simulation such that
they are  just  remembered -- instead of being parsed into ints, and
 then are written into the CSV file  as strings.

This is still work in progress
A series of test simulation have been added here. From there solutions can be easily written.
intended to fix #446

### DIFF
--- a/simul/monitor/monitor_test.go
+++ b/simul/monitor/monitor_test.go
@@ -89,38 +89,39 @@ func setupMonitor(t *testing.T) (*Monitor, *Stats) {
 	return mon, stat
 }
 
-type testStringInput map[string]string // for testing string input  in setupMonitor
-
-//type Pair struct { // this holds pairs of out put mon and stats below
-//	m *Monitor
-
-//	s *Stats
-//}
+type testStringInput map[string]string // holds the string map in setupMonitorStringTest below
 
 // setupMonitorStringTest launches a basic monitor with a created Stats object
 // When finished with the monitor, just call `End()`
-// It mimics the setMonitor above but with interface as ouput. This enables testing multiple
-// string as input
+// It mimics the setMonitor  above but uses int test count to select
+// the map to be used as input.
 func setupMonitorStringTest(t *testing.T, testCount int) (*Monitor, *Stats) {
 
-	var myMapSlice []testStringInput // for storing an array for maps. To be use for testing
+	var myMapSlice []testStringInput // for storing an array for maps.
 
+	// The map data type to be tested. At each call the testcount
+	// variable determines which one is selected
 	m1 := testStringInput{"server1": "crazyStrings"}
 	m2 := testStringInput{"server2": ""}
 	m3 := testStringInput{"server3": "123456789"}
 	m4 := testStringInput{"server4": "crazyString098765432"}
 	m5 := testStringInput{"server5": "456712309crazyString"}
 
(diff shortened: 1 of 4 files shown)


simul/monitor/stats_test.go
```diff
@@ -212,37 +213,44 @@ func TestStatsString(t *testing.T) {
...
-	for testCount, _ := range testRange {
-		mon, _ := setupMonitorStringTest(t, testCount)
-		//mon := monStatValues.m
-		dm := &DummyCounterIO{0, 0}
-		// create the counter measure
-		cm := NewCounterIOMeasure("dummy", dm)
-		if cm.baseRx != dm.rvalue || cm.baseTx != dm.wvalue {
-			t.Logf("baseRx = %d vs rvalue = %d || baseTx = %d vs wvalue = %d", cm.baseRx, dm.rvalue, cm.baseTx, dm.wvalue)
-			t.Fatal("Tx() / Rx() not working ?")
+	testRange := [5]int{1, 2, 3, 4, 5}
```
jeffallen: Before this test was covering 3 cases, now you are covering only one case, and each time the test is run it's a different case. This is not how unit tests work: they should cover the exact same set of cases on every run, so that you can compare code coverage stats from one run to the next.

simul/monitor/monitor_test.go
```diff
@@ -108,8 +108,10 @@ func setupMonitorStringTest(t *testing.T, testCount int) (*Monitor, *Stats) {
 	m1 := testStringInput{"server1": "crazyStrings"}
 	m2 := testStringInput{"server2": ""}
 	m3 := testStringInput{"server3": "123456789"}
+	m4 := testStringInput{"server3": "crazyString098765432"}
+	m5 := testStringInput{"server3": "456712309crazyString"}
```
jeffallen: This is where you put the new input in, which exercises the new desired behaviour of the code under test. Good.

But I do not see the place where you say, "the output from printing the monitor is X, and I've confirmed that the strings I expected to be in there are there".

simul/monitor/stats_test.go
```diff
@@ -212,37 +213,44 @@ func TestStatsString(t *testing.T) {
...
 	}
 
+	// Important otherwise data don't get written down to the monitor yet.
+	time.Sleep(100 * time.Millisecond)
+	str := new(bytes.Buffer)
+	stat := mon.stats
+	stat.Collect()
+	stat.WriteHeader(str)
+	stat.WriteValues(str)
+	EndAndCleanup()
```
jeffallen: I was expecting to see this code parse (even in a minimal way, i.e. "strings.Contain") str and find out if the strings that you expect to be in there are there.

jeffallen:
This was implemented in #461.

PR #460, opened 2018-08-08
Adding ip2ip tunnel and cleaning up

ineiti:
Changes the start.py script to set up ip2ip tunnels between all hosts,
so that it also works when the iccluster-servers are on different
subnets.

In addition, there are some small cleanups and specifically the setting
of the maximum numbers of file descriptors that makes bigger simulations
passing on MacOSX.

### DIFF
--- a/simul/platform/mininet.go
+++ b/simul/platform/mininet.go
@@ -5,6 +5,7 @@
 package platform
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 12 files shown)


simul/platform/mininet.go
```diff
@@ -347,7 +346,10 @@ func (m *MiniNet) parseServers() error {
 		if len(h) > 0 {
 			ips, err := net.LookupIP(h)
 			if err != nil {
-				return err
+				if err2 := CheckOutOfFileDescriptors(); err2 != nil {
```
kc1212: nitpick: I don't think you need `err2`. What's defined inside the `if` statement won't affect the outer scope.

kc1212:
I've never worked with the python file or IP tunnels so I don't fully understand what's going on. Other than than, the rest look fine, except one nitpick.

simul/platform/mininet.go
ineiti: yes, but I think it's not readable if you use the same. IIRC, this will even be an error with the linter where he'll say: "shadowing outer variable" or such.

ineiti:
Sounds like a good addition to the MININET.md... But that's for later ;)

PR #462, opened 2018-08-14
Adding libtest-documentation

ineiti:
For better understanding how to use libtest.sh, here is a documentation
pointing out the structure and the test-methods available to be
used.

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -1,0 +1,212 @@
+Navigation: [DEDIS](https://github.com/dedis/doc/tree/master/README.md) ::
+[Onet](../README.md) ::
+App support
+
+# App support
+
+This is a collection of libraries to support CLI-apps that work with the onet
+framework. The go-library simplifies loading and saving of group-configurations
+and supports some very basic input/output functions.
+
+For integration support to test the CLI-apps, `libtest.sh` supports a range
+of tests to make sure that the binary behaves as needed.
+
+# LibTest.sh
+
+This is a specialized bash-library to handle the following parts of the test:
+
+* compiling of a conode with the required services
+* setting up and removing the directory structure to run the tests
+* different test primitives to test failure or success of running the binary
+
+One goal of `libtest.sh` is to test the binary in a way as close to what a user
+would do when running the binary. This leads to a better understanding whether
+the binary has a logical setup of the commands and command line options.
+
+## Command Line Options
+
+The default for `libtest.sh`-enabled test is to:
+
+* compile the binary
+* run all the tests in a temporary directory
+* delete the temporary directory after the test.
+
+Two command line options are available to infulence how the test is run:
+
+* `-nt` - NoTemp - sets up the test-environment in `./build` and doesn't delete
+it after the test is done. This is useful if you want to inspect the environment
+after the test run. *CAREFUL* because it will *NOT* recompile the binary. You
+will use this flag mostly while developing the `test.sh` file, so that you can
+quickly test if your new tests are correct or not.
+* `-b` - Build - only useful with `-nt` to force a build of the binary. While
+developing the tests, in a first time you might want to use `-nt` until your
+`test.sh` file is OK, then you will work on your go-code. To recompile the go-code,
+use this flag. Once the go-code is OK, you can remove this flag again and work
+on the `test.sh` file.
+
+## Test.sh setup
+
+The common name for using `libtest.sh` is a file called `test.sh` in the
+same directory as the CLI-binary. The usual structure for this file is:
+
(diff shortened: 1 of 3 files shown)


app/README.md
```diff
@@ -0,0 +1,221 @@
...
+### Global Variables and inclusion of `libtest.sh`
+
+| Variable | Default | Explanation |
+| -------- | ------- | ----------- |
+| `DBG_TEST` | `0` | Show the output of the commands: 0=none, 1=test-names, 2=all |
+| `DBG_SRV` | `0` | DBG-level for server, passed as `-debug` to the `conode` command |
+| `NBR` | `3` | Highest number of servers and clients |
+| `NBR_SERVERS` | `NBR` | How many servers to configure |
+| `NBR_SERVERS_GROUP` | `NBR_SERVERS - 1` | How many servers to write to the
+group-file - per default, keep one server out of the group, to test what
```
jeffallen: This doesn't format correctly, prob needs to be one giant line. Same for testGrep, testNGRep, testCount, etc.

app/README.md
```diff
@@ -0,0 +1,221 @@
...
+To have the most important method at the top (for easier editing), `main`
+is defined at the beginning. So any edits to tests can be done there.
+
+A typical main looks like this:
+
+```bash
+main(){
+    startTest
+    buildConode
+    test Build
```
jeffallen: I find it confusing/disconcerting that test, which is normally a bash function is overridden by libTest to be a function.

I also find it unnecessarily magical that "test Build" runs some other function (testBuild). I know it doesn't seem like much difference, but the principle of least surprise would be better respected if:
* test -> runTest
* runTest runs the testing function with it's name unchanged, i.e. `runTest testBuild`

I understand that "once you get used to it" it's no big deal. But these are the kinds of confusing time wasters that turned me off libtest.sh the first time I went digging looking for what code was run for a given test. Since "jump to definition" does not work in Bash, you have to be able to keep it all in your head to know what to search for to find the code that will be run.

app/README.md
```diff
@@ -0,0 +1,221 @@
...
+    test Network
+    stopTest
+}
+```
+
+* `startTest` cleans the test-directory and builds the CLI binary
+* `buildConode` creates a new conode and automatically includes the
```
jeffallen: I remember this giving me no end of troubles to get and then to keep right. But I do think that the worst of the corner cases are gone now, so I'll hack on this when/if I still see problems. It is certainly part of the magic that made me feel like libTest was not saving me time...

app/README.md
```diff
@@ -0,0 +1,221 @@
...
+A typical main looks like this:
+
+```bash
+main(){
+    startTest
+    buildConode
+    test Build
+    test Network
+    stopTest
+}
```
kc1212: This way of writing tests is fallible, because you might forget to include your new test in `main`. Might be better to use a framework for this - see https://github.com/dedis/cothority/pull/1400

app/README.md
```diff
@@ -0,0 +1,221 @@
...
+| `testNGrep` | `String` `$@[1..]` | Asserts that `String` is NOT in the output of the
+command being run by `dbgRun` and all but the first input argument. Ignores
+the exit-code of the command. |
+| `testReGrep` | `String` | Asserts `String` is part of the last command being run
+by `testGrep` or `testNGrep`. |
+| `testReNGrep` | `String` | Asserts `String` is NOT part of the last command being run
+by `testGrep` or `testNGrep`. |
+| `testCount` | `Count` `String` `$@[2..]` | Asserts that `String` exists exactly
+`Count` times in the output of the command being run by `dbgRun` and all but
+the first two arguments. |
```
kc1212: It's easy to forget to update these, so I feel these should go into the code comments first.

app/README.md
```diff
@@ -0,0 +1,221 @@
...
+    runCoBG 1 2
+    testOut "Running network"
+    testGrep "Available_Services" runCl -g public.toml
+    testGrep "Available_Services" runCl -g public.toml
+}
+```
+
+* `runCoBG 1 2` starts conode number `1` and `2` (the index starts at 1) in
+the background. There is some code to make sure the conodes are actually up
+and running and listening on the websocket-ports.
```
kc1212: Is there a reason why we don't have `runCoBG 2` to mean start 2 conodes, similar to `run_conode.sh`? Is it because we define the number of conodes in `$NBR`? This function confused me in the past.

app/README.md
kc1212: made an issue for it https://github.com/dedis/onet/issues/463

PR #464, opened 2018-08-15
Add versions to db

ineiti:
This PR adds LoadVersion and SaveVersion to store a db-version int for every service.
It also adds LoadRaw to fetch old versions of the db.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,32 +1,49 @@
 package onet
 
 import (
+	"bytes"
+	"encoding/binary"
 	"fmt"
 	"sync"
 
 	bolt "github.com/coreos/bbolt"
+	"github.com/dedis/onet/log"
 	"github.com/dedis/onet/network"
 )
 
 // Context represents the methods that are available to a service.
 type Context struct {
-	overlay    *Overlay
-	server     *Server
-	serviceID  ServiceID
-	manager    *serviceManager
-	bucketName []byte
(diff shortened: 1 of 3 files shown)


ineiti:
Hmm - how could this happen? You mean that we do something where we cannot update? Then the service would have to `panic` and tell the user that he cannot use this database. Or he would have to wipe out the data (perhaps create a backup first) and then start from scratch.

Or did I miss something here?

kc1212:
If I have a very old database, suppose it's version 0.1.0, then I upgrade my code to version 1.1.0. The code will refuse to run just by looking at the version numbers and tell the user that its database is too old and there is not upgrade path. If the new version is 0.2.0, then the code understands that an upgrade is possible and should have to code to do so.

What I described above is certainly possible with integer versions. But you'll need to store the version numbers that the update supports, so it's a bit less obvious than using semantic versioning. However, if the plan is to provide upgrade paths for all versions of the database, then I agree semantic versioning is not needed.

ineiti:
So I vote against semantic versioning. Save this thread for when we want to update `v2` and change the version, but this version number is already taken by `v3` and blame me...

kc1212:
With this new setting in GitHub, do I have to "Approve" the PR in order to merge? It seems I cannot change my old review from "Comment" to "Approve", maybe @jeffallen has to add a review...

PR #466, opened 2018-08-21
change non-working public ip detetcion service to 3rd party working one

yurunhub:
https://dedis.ch/portscan.php does not seem to work so changed to 3rd public one https://api.ipify.org?format=text

### DIFF
--- a/app/server.go
+++ b/app/server.go
@@ -33,7 +33,7 @@ const DefaultPort = 6879
 const DefaultAddress = "127.0.0.1"
 
 // Service used to get the public IP-address.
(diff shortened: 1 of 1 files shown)


jeffallen:
We made a change a while back so that dedis.ch redirects to dedis.epfl.ch. However, blog.dedis.ch still runs out of the same directory, and https://blog.dedis.ch/portscan.php does what https://dedis.ch/portscan.php used to do. So please update this PR to use that URL, and then I'll merge it. Thanks.

yurunhub:
new URL tested and corrected

PR #469, opened 2018-09-04
Add timeouts for dialling new connections and sending

### LINKED ISSUE #468: Send() and net.Dial called with no timeout
When looking into how SendRaw() might behave in the face of failed servers, I noticed that we are calling net.Dial with no configuration (see NewTCPConn and NewTLSConn, which calls tls.Dial instead of tls.DialWithDialer.

That means we are depending on the OS timeout, which is likely way too long. To investigate and choose a way forward.

Also: @kc1212 noted: "there needs to be a version of SendRaw that supports a timeout"

jeffallen:
Fixes #468.

### DIFF
--- a/network/local.go
+++ b/network/local.go
@@ -270,7 +270,7 @@ func (lc *LocalConn) start(wg *sync.WaitGroup) {
 // Send takes a context (that is not used in any way) and a message that
 // will be sent to the remote endpoint.
 // If there is an error in the connection, it will be returned.
-func (lc *LocalConn) Send(msg Message) (uint64, error) {
+func (lc *LocalConn) Send(msg Message, opt ...Option) (uint64, error) {
 	buff, err := Marshal(msg)
 	if err != nil {
 		return 0, err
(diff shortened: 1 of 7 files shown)


jeffallen:
Hmm, so this code is right, but not quite right. The question is, which callers will decide to set timeouts, and how? Because the code that got us thinking about this was the SendRaw in omniledger/service/viewchange.go, where we know we are almost certainly trying to send to the ex-leader, who has demonstrated an unwillingness to talk on the network, leading to the view change.

So, should SendRaw take Options as well? That would require adding it into a few more places, but nothing too ugly. Or should SendRaw have a default timeout built into it, and always use that?

ineiti:
Not sure what is right here - we might put a default timeout in the `Context` and use that. But I see this call-chain:
```
Context.SendRaw ->
Router.Send ->
Conn.Send
```

And if I followed correctly, only `Conn.Send` has the timeout option. So you have to add the timeout option at least to the `Router.Send`, and either to `Context.SendRaw` OR use a default, user changeable timeout stored in `Context`.

jeffallen:
Littering the onet api with timeouts is not a good plan. The problem we're worried about is real: Go seems to make no guarantees whatsoever about when net.Dial will return in the absence of timeouts, and with MaxRetryConnect of 5, even if we added a hardcoded timeout on net.Dial of 1 minute, viewchange's loop to do a SendRaw to every server in the roster might take a long time.

The simplest solution is to rename the existing global readTimeout into networkTimeout, and use it for net.Dial, tls.Dial, and for tcp.go's sendRaw (local.go does not have timeout risks, and tls.go leverages tcp.go's sendRaw).

But that will still leave view change risking blocking, so we'll change it to send those via goroutines.

PR #470, opened 2018-09-05
Apply the same timeout for reads to send and connect too

### LINKED ISSUE #468: Send() and net.Dial called with no timeout

jeffallen:
There was an existing timeout for reads. Rename it and then use it to prevent
sends from taking too long, with/without the need to connect.

Fixes #468.

### DIFF
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -14,14 +14,14 @@ import (
 	"github.com/dedis/onet/log"
 )
 
-// a connection will return an io.EOF after readTimeout if nothing has been
-// sent.
-var readTimeout = 1 * time.Minute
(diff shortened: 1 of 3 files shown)


kc1212:
Is the failure related to flaky tests? I just merged https://github.com/dedis/onet/pull/474, could you rebase and check if it passes?

PR #473, opened 2018-09-06
Run TestClose in parallel

kc1212:
We run TestClose in parallel because if one of the TestClose takes a
long time then it may cause strange behaviour in other services.

### DIFF
--- a/server.go
+++ b/server.go
@@ -162,12 +162,18 @@ func (c *Server) Close() error {
 	// For all services that have `TestClose` defined, call it to make
 	// sure they are able to clean up. This should only be used for tests!
 	c.serviceManager.servicesMutex.Lock()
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -162,12 +162,18 @@ func (c *Server) Close() error {
...
 	// sure they are able to clean up. This should only be used for tests!
 	c.serviceManager.servicesMutex.Lock()
+	var wg sync.WaitGroup
 	for _, serv := range c.serviceManager.services {
-		s, ok := serv.(TestClose)
-		if ok {
-			s.TestClose()
-		}
+		wg.Add(1)
+		func(s Service) {
```
ineiti: missing `go` to make this parallel, no?

PR #478, opened 2018-09-19
Adding streaming abilities to onet websockets

kc1212:
Adding streaming abilities to onet websockets.

We introduce a way to register handlers - `RegisterStreamingHandler`. This handler will return a channel to stream data to the client upon receiving a message. Internally, the processor listens on the channel and forwards all the messages to the websocket loop which then forwards them to the client.

If the client closes its connection. The websocket loop sends a stop signal back up the path to tell the service to stop streaming.

There shouldn't be any backward incompatible changes, `make test` works on cothority master.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -45,15 +45,21 @@ var errType = reflect.TypeOf((*error)(nil)).Elem()
 //
 //  * msg is a pointer to a structure to the message sent.
 //  * ret is a pointer to a struct of the return-message.
-//  * err is can be nil, or any type that implements error.
+//  * err is an error, it can be nil, or any type that implements error.
 //
 // struct_name is stripped of its package-name, so a structure like
 // network.Body will be converted to Body.
 func (p *ServiceProcessor) RegisterHandler(f interface{}) error {
-	if err := p.handlerSanityCheck(f); err != nil {
+	if err := p.handlerInputCheck(f); err != nil {
 		return err
 	}
+
+	// check output
 	ft := reflect.TypeOf(f)
+	if ft.NumOut() != 2 {
+		return errors.New("Need 2 return values: network.Body and error")
+	}
+	// first output
 	ret := ft.Out(0)
 	if ret.Kind() != reflect.Interface {
 		if ret.Kind() != reflect.Ptr {
@@ -62,6 +68,11 @@ func (p *ServiceProcessor) RegisterHandler(f interface{}) error {
 		if ret.Elem().Kind() != reflect.Struct {
 			return errors.New("1st return value must be a pointer to a *struct* or an interface")
 		}
+	}
+	// second output
+	if !ft.Out(1).Implements(errType) {
+		return errors.New("2nd return value has to implement error, but is: " +
+			ft.Out(1).String())
 	}
 
 	cr := ft.In(0)
@@ -72,18 +83,58 @@ func (p *ServiceProcessor) RegisterHandler(f interface{}) error {
 	return nil
 }
 
-// RegisterStreamingHandler TODO document
-// 2. func(msg interface{})(ret chan interface{}, err, error)
+// RegisterStreamingHandler stores a handler that is responsible for streaming
+// messages to the client via a channel. Websocket will accept requests for
+// this handler at "ws://service_name/struct_name", where struct_name is
+// argument of f, which must be in the form:
+// func(msg interface{})(retChan chan interface{}, closeChan chan bool, err error)
+//
+//  * msg is a pointer to a structure to the message sent.
+//  * retChan is a channel of a pointer to a struct, everything sent into this
+//    channel will be forwarded to the client, if there are no more messages,
+//    the service should close retChan.
+//  * closeChan is a boolean channel, upon receiving a message on this channel,
+//    the handler should stop sending messages and close retChan.
+//  * err is an error, it can be nil, or any type that implements error.
+//
+// struct_name is stripped of its package-name, so a structure like
+// network.Body will be converted to Body.
 func (p *ServiceProcessor) RegisterStreamingHandler(f interface{}) error {
-	if err := p.handlerSanityCheck(f); err != nil {
+	if err := p.handlerInputCheck(f); err != nil {
 		return err
 	}
+
+	// check output
 	ft := reflect.TypeOf(f)
-	ret := ft.Out(0)
-	if ret.Kind() != reflect.Chan {
+	if ft.NumOut() != 3 {
+		return errors.New("Need 3 return values: chan interface{}, chan bool and error")
+	}
+	// first output
+	ret0 := ft.Out(0)
+	if ret0.Kind() != reflect.Chan {
 		return errors.New("1st return value must be a channel")
 	}
-	// TODO is there a way to check the kind inside the channel?
+	if ret0.Elem().Kind() != reflect.Interface {
+		if ret0.Elem().Kind() != reflect.Ptr {
+			return errors.New("1st return value must be a channel of a *pointer* to a struct")
+		}
+		if ret0.Elem().Elem().Kind() != reflect.Struct {
+			return errors.New("1st return value must be a channel of a pointer to a *struct*")
+		}
+	}
+	// second output
+	ret1 := ft.Out(1)
+	if ret1.Kind() != reflect.Chan {
+		return errors.New("2nd return value must be a channel")
+	}
+	if ret1.Elem().Kind() != reflect.Bool {
+		return errors.New("2nd return value must be a boolean channel")
+	}
+	// third output
+	if !ft.Out(2).Implements(errType) {
+		return errors.New("3rd return value has to implement error, but is: " +
+			ft.Out(2).String())
+	}
 
 	cr := ft.In(0)
 	log.Lvl4("Registering streaming handler", cr.String())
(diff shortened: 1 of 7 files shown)


processor.go
```diff
@@ -39,8 +39,9 @@ var errType = reflect.TypeOf((*error)(nil)).Elem()
 
 // RegisterHandler will store the given handler that will be used by the service.
 // WebSocket will then forward requests to "ws://service_name/struct_name"
-// to the given function f, which must be of the following form:
-// func(msg interface{})(ret interface{}, err error)
+// to the given function f, which must be one of the following forms:
+// 1. func(msg interface{})(ret interface{}, err error)
+// 2. func(msg interface{})(ret chan interface{}, err, error)
```
ineiti: I wonder whether we should put both in the same `RegisterHandler` or if we should not have `RegisterHandler` and `RegisterStream`. I have a small preference for the latter.

processor.go
```diff
@@ -70,14 +71,20 @@ func (p *ServiceProcessor) RegisterHandler(f interface{}) error {
...
-		if ret.Kind() != reflect.Ptr {
-			return errors.New("1st return value must be a *pointer* to a struct or an interface")
-		}
-		if ret.Elem().Kind() != reflect.Struct {
+		if ret.Kind() == reflect.Ptr {
+			if ret.Elem().Kind() != reflect.Struct {
+				return errors.New("1st return value must be a pointer to a *struct* or an interface")
+			}
+		} else if ret.Kind() == reflect.Chan {
+			// TODO is there a way to check the kind inside the channel?
```
ineiti: Look at `RegisterChannels` - I think there is something in there.

processor.go
```diff
@@ -136,19 +150,51 @@ func (p *ServiceProcessor) ProcessClientRequest(req *http.Request, path string,
...
 			return nil, ierr.(error)
 		}
 		return ret[0].Interface(), nil
 	}()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
+
+	if reply != nil && reflect.TypeOf(reply).Kind() == reflect.Chan {
```
ineiti: This would be cleaner if we had different handlers for simple requests and streams.

websocket.go
```diff
@@ -193,15 +193,30 @@ func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			if replyChan == nil {
+				tx += len(reply)
+				err := ws.WriteMessage(mt, reply)
+				if err != nil {
+					log.Error(err)
+					return
+				}
+			} else {
+				go func() {
+					// TODO fix
```
ineiti: Does the fix include listening on the `closing` channel here?

websocket.go
```diff
@@ -268,6 +283,8 @@ func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]b
 	dest := destination{dst, path}
 	conn, ok := c.connections[dest]
 	if !ok {
+		// TODO we are opening a new connection for every new path?
+		// not possible to use an existing connection for the same service?
```
ineiti: Might be possible. When I started the websockets, I think I missed a lot of points and should've gone with a REST interface over http(s)...

websocket.go
```diff
@@ -193,15 +193,30 @@ func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-				return
+			if replyChan == nil {
+				tx += len(reply)
+				err := ws.WriteMessage(mt, reply)
+				if err != nil {
+					log.Error(err)
+					return
+				}
+			} else {
+				go func() {
```
jeffallen: Whacky. So if a client issues 2 requests on one websocket channel, and both have a streaming return, then the results from the two requests will arrive on the websocket, intermingled. And probably the call to ws.WriteMessage is racy.

In other words: I'm not sure about that `go`. :)

I was expecting that a ws connection used for streaming would be dedicated to delivering results until closed. (Not loop back to read again.)

ineiti:
Tried out a simple playground to reduce reflect usage, but it fails:

https://play.golang.org/p/YCRjNbwCxky

Do I miss something obvious?

processor.go
```diff
@@ -136,19 +150,51 @@ func (p *ServiceProcessor) ProcessClientRequest(req *http.Request, path string,
...
+					if err != nil {
+						log.Error(err)
+						// TODO close these?
+						close(outChan)
+						case0.Chan.Close()
+						return
+					}
+					outChan <- buf
+				}
+				// We don't add a way to close here, otherwise
```
jeffallen: Yes, these issues are going to be the hardest. Starting streaming is easy, knowing when to stop is hard.

We need that an error on write down towards the ws client results in closing the channel, so that the service can stop sending.

And we'll need to call SetWriteDeadline on the ws, so that if the client forgets to read from us, it cannot hold open the server-side resources.

websocket.go
```diff
@@ -193,15 +193,30 @@ func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		s := t.service
 		var reply []byte
+		var replyChan chan []byte
 		path := strings.TrimPrefix(r.URL.Path, "/"+t.serviceName+"/")
 		log.Lvlf2("ws request from %s: %s/%s", r.RemoteAddr, t.serviceName, path)
-		reply, err = s.ProcessClientRequest(r, path, buf)
+		reply, replyChan, err = s.ProcessClientRequest(r, path, buf)
```
jeffallen: I think I was expecting ProcessClientRequest to return a reply as usual, and then optionally a channel of replies.

So a client issues a search and gets a reply with 10 hits. Or a client issues a search with req.Follow = True, and then they get 10 hits and also a reply channel streaming new hits to them.

kc1212:
Regarding https://github.com/dedis/onet/pull/478#issuecomment-422828889

The way I understood it is that if Cat is a subclass of Animal. Then you can call the Animal methods on Cat. But, if you have `List<Cat>` and `List<Animal>`, and you implement some methods on `List<Animal>`, you cannot use those methods on `List<Cat>`. I think `func(x interface{})` vs `func(i int)` is the same story, once you wrap things that have relationships inside something else, then you lose that relationship.

Scala has a solution for this called [variances](https://docs.scala-lang.org/tour/variances.html), Java probably has it too.

kc1212:
Regarding https://github.com/dedis/onet/pull/478#pullrequestreview-156848383

The way we're doing things now need reflections for reasons I mentioned above. The other possibility I can think of is to use an interface like:
```
func RegisterHandler(f Handler) error
type Handler interface {
  func Call(in interface{}) (out interface{}, error)
  func GetPath() string
}
```
but it'll introduce a lot of repeated code which you have to write to cast the incoming message to the right type and computing the path from the message type.

kc1212:
Okay, I'll try to do that today.

service_test.go
```diff
@@ -615,13 +615,13 @@ func (ds *DummyService) ProcessClientRequest(req *http.Request, path string, buf
 
 	if err := ds.c.RegisterProtocolInstance(dp); err != nil {
 		ds.link <- false
-		return nil, errors.New("")
+		return nil, nil, errors.New("")
```
jeffallen: What is this errors.New("") about? (I know it was there before, but still seems strange.)

websocket.go
```diff
@@ -193,25 +193,45 @@ func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				for {
+					select {
+					case reply, ok := <-tun.out:
+						if !ok {
+							err = errors.New("service finished streaming")
+							close(tun.close)
+							break outerReadLoop
+						}
+						tx += len(reply)
+						err = ws.WriteMessage(mt, reply)
```
jeffallen: I was expecting to see something about Write deadlines here.

websocket.go
```diff
@@ -182,6 +181,7 @@ func (t wsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}()
 
 	// Loop for each message
+outerReadLoop:
 	for err == nil {
 		mt, buf, rerr := ws.ReadMessage()
```
jeffallen: I thought there was some Read deadline here. Perhaps this is a follow-on work item...

websocket_test.go
```diff
@@ -458,6 +458,198 @@ func TestWebSocketTLS_Error(t *testing.T) {
 	require.NotEqual(t, "", log.GetStdOut())
 }
 
+// TestWebSocket_Streaming performs 3 test cases.
```
jeffallen: This test is 💯 👍

websocket.go
kc1212: I see that gorilla websockets has `Set{Read,Write}Deadline`, I could use that. But what would be the appropriate deadline, a minute? Assuming it's to stop client/server from sending unrealistically large messages.

service_test.go
kc1212: I'm not sure, but it makes more sense to return `err`.

kc1212:
We could have `StreamingTunnel` start a go-routine that pipe messages internally and close the external channels if no messages are received within some time duration. But it'll be difficult to set this duration because some services might very well be quite during off-peak hours (e.g., weekends) and you don't want to close client connections if the client is acting more like a server that's doing some post-processing.

websocket.go
jeffallen: It is to stop a badly behaved client from trickling a message in a byte at a time, holding open server resources. Same for reading... for example when a server has a torrent of messages and the client just stops talking TCP (phone out of GSM range, out of battery, etc) then all those messages sit in kernel buffers until TCP times out.

ineiti:
Sounds good to go forward and later (probably never) have 5 minutes timeout.

PR #487, opened 2018-10-24
LRU cache with maximum size onet-273

### LINKED ISSUE #273: Rate control on roster creation
Any given server-to-server connection should be rate-limited in it's ability to consume resources (which resources? Rosters? Trees?) on the target server.

Add expiration logic to Roster cache.
Add expiration logic to Tree cache.

Gilthoniel:
This PR changes the cache to use a LRU policy with a maximum number of entries. This maximum is for the moment static when instantiating the cache.

It also removes the goroutine that cleans the expired items by a simple check when getting elements as we can simply check the tail expiration.

Fixes #273

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/dedis/onet/log"
 	uuid "github.com/satori/go.uuid"
 )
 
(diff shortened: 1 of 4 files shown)


jeffallen:
Please write commit messages like other ones already in use in the team. It should have first line which is a title, and then a blank line, then a few sentences explaining WHY (the what is the code; they why is in the documentation). And then it should finish with the right syntax to auto-close the Github issue. https://blog.github.com/2013-01-22-closing-issues-via-commit-messages/

Gilthoniel:
Duly noted. Its just a bad habit from my previous job but you are totally right.

PR #494, opened 2018-11-09
Fix TreeNodeInstance rx status

### LINKED ISSUE #481: Rx is always zero in TreeNodeInstance
Doing `n.rx.add(uint64(len(onetMsg.MsgSlice)))` doesn't work, the slice is always zero. Maybe `Overlay:Process` needs to be fixed to include `MsgSlice`. It's not used at the moment:
```
		protoMsg := &ProtocolMsg{
			From:           info.TreeNodeInfo.From,
			To:             info.TreeNodeInfo.To,
			ServerIdentity: env.ServerIdentity,
			Msg:            inner,
			MsgType:        typ,
		}
```
This issue was discovered while debugging monitoring problems with @cba33 in onet.v2, but the problem exists in master too.

Gilthoniel:
This fixes the rx status that was always 0 even
after processing a message.

Fixes #481

### DIFF
--- a/treenode_test.go
+++ b/treenode_test.go
@@ -13,6 +13,7 @@ import (
 
 func init() {
 	GlobalProtocolRegister(spawnName, newSpawnProto)
+	GlobalProtocolRegister(dummyName, newDummyProto)
 }
 
 func TestTreeNodeCreateProtocol(t *testing.T) {
@@ -32,6 +33,21 @@ func TestTreeNodeCreateProtocol(t *testing.T) {
 	<-spawnCh
 
 	pi.(*spawnProto).Done()
+}
+
+func TestTreeNodeRxTx(t *testing.T) {
+	local := NewLocalTest(tSuite)
+	defer local.CloseAll()
+
+	_, _, tree := local.GenTree(2, true)
+	pi, err := local.StartProtocol(dummyName, tree)
+	require.Nil(t, err)
+	protocol := pi.(*dummyProto)
+
+	<-dummyDone
+	<-dummyDone
+	require.NotZero(t, protocol.TreeNodeInstance.Rx())
+	require.NotZero(t, protocol.TreeNodeInstance.Tx())
 }
 
 func TestHandlerReturn(t *testing.T) {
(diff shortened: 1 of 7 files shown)


treenode_test.go
```diff
@@ -194,3 +210,55 @@ func (s *spawnProto) HandlerError2(msg spawnMsg) error {
 func (s *spawnProto) HandlerError3(msg spawnMsg) (int, error) {
 	return 0, nil
 }
+
+// Simple protocol to have messages transmit between nodes
+const dummyName = "TestDummy"
```
kc1212: can you call this `pingPongProto` or something like that, i think that's what it does

treenode_test.go
```diff
@@ -194,3 +210,55 @@ func (s *spawnProto) HandlerError2(msg spawnMsg) error {
...
+		cp.SendToParent(&DummyProtoMsg{})
+	}
+
+	cp.Done()
+	dummyDone <- true
+	return nil
+}
+
+func (cp *dummyProto) Start() error {
+	if !cp.IsRoot() {
```
kc1212: you don't need this part, only the root will call start

treenode_test.go
```diff
@@ -194,3 +210,55 @@ func (s *spawnProto) HandlerError2(msg spawnMsg) error {
...
+type dummyProto struct {
+	*TreeNodeInstance
+	done      chan bool
+	dummyChan chan struct {
+		*TreeNode
+		DummyProtoMsg
+	}
+}
+
+var dummyDone = make(chan bool)
```
kc1212: instead of using a global channel, you can make a channel in the `dummyProto` struct, and then write to it when Dispatch is done on root.  Actually, you already have it - `done chan bool`.

Then just write to the channel in dispatch like 
```

func (cp *dummyProto) Dispatch() error {
	select {
	defer cp.Done()
	case <-cp.dummyChan:
		if !cp.IsRoot() { cp.SendToParent(&DummyProtoMsg{}) }
	}
	cp.done  <- true // probably needs to be buffered so that the children can return
	return nil
} 
```
I think that should work...

Gilthoniel: Yes, and that's much cleaner, thanks!

PR #495, opened 2018-11-14
Roster concatenation

Gilthoniel:
This adds a function that will create a new roster using an existing
one and a list of server identities.

The helper function will be use to concatenate two rosters from two different blocks when propagating forward-links so that every conode implied will know about it.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -701,6 +701,24 @@ func (ro *Roster) Contains(pubs []kyber.Point) bool {
 	return true
 }
 
+// Concat makes a new roster using an existing one and a list
+// of server identities
+func (ro *Roster) Concat(sis []*network.ServerIdentity) *Roster {
+	roMap := make(map[string]*network.ServerIdentity)
+	for _, si := range append(ro.List, sis...) {
+		roMap[si.String()] = si
+	}
+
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -701,6 +701,24 @@ func (ro *Roster) Contains(pubs []kyber.Point) bool {
...
+func (ro *Roster) Concat(sis []*network.ServerIdentity) *Roster {
+	roMap := make(map[string]*network.ServerIdentity)
+	for _, si := range append(ro.List, sis...) {
+		roMap[si.String()] = si
+	}
+
+	newList := make([]*network.ServerIdentity, len(roMap))
+	i := 0
+	for _, si := range roMap {
+		newList[i] = si
```
ineiti: you can use `make(..., ..., 0)` and then use `append` here, so you don't need the counter.

Gilthoniel: Ok, I wasn't sure which one was the most efficient

ineiti: Hmm - good question. As this is not a tight inner-loop that we're running 1000 times per second, it doesn't really matter. So I concentrate on readability, which is quite subjective...

PR #507, opened 2018-12-06
Key pair per service

Gilthoniel:
This adds service keypairs to the configuration and
add getters for convenient usage.


network/struct.go
```diff
@@ -149,6 +179,42 @@ func (si *ServerIdentity) GetPrivate() kyber.Scalar {
 	return si.private
 }
 
+// ServicePublic returns the public key of the service or the default
+// one if the service has not been registered with a suite
+func (si *ServerIdentity) ServicePublic(name string) kyber.Point {
+	for _, si := range si.ServiceIdentities {
```
jeffallen: Please don't shadow variables quite so close to each other. Suggestion: use svcid for the iterator here.

service.go
```diff
@@ -159,6 +172,23 @@ func (s *serviceFactory) registeredServiceIDs() []ServiceID {
 	return ids
 }
 
+// generateKeyPairs generates the key pairs for the services that
```
jeffallen: It is surprising to me that key generation is not in the appcfg directory. It seems like something that ONLY happens during config file creation, not automagically here in onet. (But if you can explain your reasoning, then I wouldn't be surprised and would be ok with this.)

simulation.go
```diff
@@ -68,12 +68,27 @@ type SimulationConfig struct {
 	Config string
 }
 
+// SimulationPrivateKey contains the default private key and the service
```
jeffallen: Wow, we forgot about that part of the work. Good job.

network/struct.go
```diff
@@ -149,6 +179,42 @@ func (si *ServerIdentity) GetPrivate() kyber.Scalar {
...
+		}
+	}
+
+	return si.Public
+}
+
+// ServicePrivate returns the private key of the service or the default
+// one if the service has not been registered with a suite
+func (si *ServerIdentity) ServicePrivate(name string) kyber.Scalar {
+	for _, si := range si.ServiceIdentities {
```
ineiti: same here: `s/si/svcid/`

PR #511, opened 2018-12-20
Change the default port base when generating hosts

Gilthoniel:
One recurrent problem on Jenkins is caused by multiple jobs running in parallel and causing conflicts with port binding. This PR tries to fix that issue by using a parameter that allows to change the base (currently 2000) when generating hosts.

I explored different options to even work on multithreading executions but this would require a lock in a file or similar and that would be dangerous when the cleaning routine fails.

### DIFF
--- a/local.go
+++ b/local.go
@@ -69,9 +69,6 @@ type LocalTest struct {
 	// since now the temp directory is gone.
 	closed bool
 	T      *testing.T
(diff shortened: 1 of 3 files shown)


Gilthoniel:
Closed in favor of dockerized unit tests

PR #513, opened 2019-01-23
Assert that packets' sender address is correct

ineiti:
@jeffallen discovered an attack where a node maliciously changes the  field of the
message being sent to a node, thus impersonating another node in the system.
This patch verifies that the sender is correct.

TODO: better catch the error.

### DIFF
--- a/treenode.go
+++ b/treenode.go
@@ -407,6 +407,12 @@ func (n *TreeNodeInstance) reflectCreate(t reflect.Type, msg *ProtocolMsg) refle
 			m.Field(0).Set(reflect.ValueOf(tn))
 			m.Field(1).Set(reflect.Indirect(reflect.ValueOf(msg.Msg)))
 		}
+		// Check whether the sender treenode actually is the same as the node who sent it.
+		// We can trust msg.ServerIdentity, because it is written in Router.handleConn and
+		// is not writable by the sending node.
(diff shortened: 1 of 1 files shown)


jeffallen:
I am working on a unit test to prove that this is a problem.

jeffallen:
I tried to find a way to test this, but I couldn't. Basically, it would be necessary to create the correct network.Envelope to forge the tree node and send it across a raw TCP connection, I think. In the end I decided it was better to just check this in based on the fact it passes the current tests than spend more time trying to exercise it.

treenode.go
```diff
@@ -398,7 +405,7 @@ func (n *TreeNodeInstance) dispatchHandler(msgSlice []*ProtocolMsg) error {
 	return nil
 }
 
-func (n *TreeNodeInstance) reflectCreate(t reflect.Type, msg *ProtocolMsg) reflect.Value {
+func (n *TreeNodeInstance) reflectCreate(t reflect.Type, msg *ProtocolMsg) (reflect.Value, error) {
```
kc1212: i feel it's not 100% clean to do this check inside `reflectCreate` because the function name doesn't tell the programmer what it's doing accurate. maybe it's better to change the function name to reflect its now semantics e.g., `createValueAndVerify`, or move the verification logic into a new function.

kc1212:
Are there tests to check that it work?

PR #526, opened 2019-03-11
Prevent Roster to be nil during a TreeNodeInstance life

Gilthoniel:
This PR fixes the nil pointer access issue caused by the cache by removing the different caches to keep only a tree storage (the roster is inside the tree anyway) and by keeping the tree stored until the tree node instance has finished its protocol. It also checks that another instance is not using the tree.


Gilthoniel:
Backwards compatibility tested with 4 v3.0.0 nodes and 4 of the new version

treestorage.go
```diff
@@ -0,0 +1,136 @@
...
+
+	go func() {
+		defer ts.wg.Done()
+
+		timer := time.NewTimer(ts.timeout)
+
+		select {
+		// other distant node instances of the protocol could ask for the tree even
+		// after we're done locally and then it needs to be kept around for some time
+		// TODO: could this be optimized ?!
```
jeffallen: So, could it? My vote is no, and nuke the comment.

treestorage.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Get returns the tree if it exists or nil
+func (ts *treeStorage) Get(id TreeID) *Tree {
+	ts.lock.Lock()
+	defer ts.lock.Unlock()
+
+	return ts.trees[id]
+}
+
+// Set sets the given tree adn cancel potential removal
```
jeffallen:
```suggestion
// Set sets the given tree and cancel potential removal
```

treestorage.go
```diff
@@ -0,0 +1,136 @@
...
+		// other distant node instances of the protocol could ask for the tree even
+		// after we're done locally and then it needs to be kept around for some time
+		// TODO: could this be optimized ?!
+		case <-timer.C:
+			ts.lock.Lock()
+			delete(ts.trees, id)
+			delete(ts.cancellations, id)
+			ts.lock.Unlock()
+		case <-c:
+			timer.Stop()
```
jeffallen: I'm a tiny bit concerned about what if timer.C and c become readable in a race, and select chooses c to wake up. timer.C will be a buffered channel, so the time will go into it and any goroutine assocaited with writing that will not be blocked, so that's ok. And then timer can be GC'd after the return on line 106, so I guess there's nothing to worry about.

(I just kept this comment here to remember what I investigated and decided. nothing for you to do.)

messages.go
```diff
@@ -139,21 +145,34 @@ type TreeNodeInfo struct {
...
+type RequestRoster struct {
+	RosterID RosterID
 }
 
 // RequestTree is used to ask the parent for a given Tree
 type RequestTree struct {
 	// The treeID of the tree we want
 	TreeID TreeID
+	// Version of the request tree
+	Version uint32
```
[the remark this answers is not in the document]
kc1212: Should we make a list of all these ugly backward compatibility code somewhere so that they can be removed in v4?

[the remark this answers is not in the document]
Gilthoniel: I actually put the `Deprecated: ` thing 1.) to explain and 2.) make an easy searchable tag to remove. Now it might be interesting to have a list of what else can be removed if we remove that maybe.

PR #527, opened 2019-03-13
Cd into modules directory to do go build

kc1212:
Fixes the deterlab part of dedis/cothority#1763

Tested on deterlab with normal go.mod file and `redirect`.

### DIFF
--- a/simul/platform/cliutils.go
+++ b/simul/platform/cliutils.go
@@ -5,12 +5,10 @@ package platform
 import (
 	"bufio"
 	"bytes"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strings"
 
 	"go.dedis.ch/onet/v3/log"
 )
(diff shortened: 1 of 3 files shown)


simul/platform/cliutils.go
```diff
@@ -98,11 +100,9 @@ func Build(path, out, goarch, goos string, buildArgs ...string) (string, error)
...
 	wd, _ := os.Getwd()
 	log.Lvl4("In directory", wd)
-	var args []string
-	args = append(args, "build", "-v")
-	args = append(args, buildArgs...)
-	args = append(args, "-o", out, path)
-	cmd = exec.Command("go", args...)
+	// we have to cd into the directory to do the build when using go
+	// modules, not sure about the exact reason for this behaviour yet
+	cmd = exec.Command("/bin/bash", "-c", fmt.Sprintf("(cd %s && go build -v %s -o %s)", path, strings.Join(buildArgs, " "), out))
```
jeffallen: I have seen students launch simulations from Windows, and this will fail because it depends on /bin/bash. Instead, please use https://godoc.org/os/exec#Cmd.Dir instead to set the working directory.

PR #528, opened 2019-03-19
Fix the roster ID generation

Gilthoniel:
This sorts the service identities because it can happen the order
array change when writing/reading toml files. This makes sure it
doesn't impact the final ID.

This issue occurs when checking the roster sent back when creating a genesis block in dedis/cothority#1764

### DIFF
--- a/network/struct.go
+++ b/network/struct.go
@@ -120,6 +120,21 @@ func NewServiceIdentityFromPair(name string, suite suites.Suite, kp *key.Pair) S
 	return NewServiceIdentity(name, suite, kp.Public, kp.Private)
 }
 
+// ServiceIdentities provides definitions to sort the array by service name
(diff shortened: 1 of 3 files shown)


Gilthoniel:
It would be necessary to increase the patch version to use the fix in cothority

jeffallen:
Can you explain the backwards compatibility impacts of this again better in your comment above?

Gilthoniel:
So the first backwards compatibility was a bit shaky because the ID is usually transferred through protobuf so you don't recalculate it. But when you derive the roster you do. The order is not supposed to have changed because it happens when you read/write toml files, that's why we never had this issue. But to have 100% certainty of backwards compatibility, I updated the code so that we can gradually use the fix.

PR #529, opened 2019-03-20
Service Aggregate calculation

ineiti:
Adds an aggregate calculation for the Roster and the TreeNodeInstance, which
comes in handy for drynx.


ineiti:
Good point! In fact it's even worse: As I'm cycling through `ServicePublics`, I can have multiple groups of points in case of a missing service public key!

I propose to change `ServerIdentity.ServicePublic` to either return `nil`, or an `error` in case the specific key for the service doesn't exist. Then we can catch this error in the `ServicePublics` and `ServiceAggregate`.

ineiti:
After discussing with @Gilthoniel , we concluded that I should not change the signatures of the methods. So I only change the `ServiceAggregate` of the `Roster` and the `Aggregate` of the `TreeNodeInstance`...

PR #530, opened 2019-03-25
Register bn256.G1 and .GT

ineiti:
Previously only G2 has been registered to dedis/protobuf.
Now also G1 and GT are registered.

Depends on dedis/kyber#370


go.mod
```diff
@@ -22,3 +22,5 @@ require (
 	gopkg.in/tylerb/graceful.v1 v1.2.15
 	rsc.io/goversion v1.2.0
 )
+
+replace go.dedis.ch/kyber/v3 => ../kyber
```
Gilthoniel: this must go away

ineiti: Yes, as soon as kyber has a new tag ;)

PR #535, opened 2019-04-05
Fix leaking goroutine in treestorage

Gilthoniel:
After the server is closed, it can still process messages and this triggers the storage to clean some trees which creates again new goroutines that won't be closed. This addresses the scenario.

Note: this fixes a flaky test in dedis/cothority


Gilthoniel:
Note to myself: if Github is asking to update the branch, don't stupidly click the button

PR #537, opened 2019-04-29
Fix data race when getting new buckets

Gilthoniel:
It might happen that the append builtin was modifying the inner array and then creating a data race. The modification should not happen anyway so this is fixed by copying the array so that there is only reads.


Gilthoniel:
Required by dedis/cothority#1828

kc1212:
Go ahead and merge if there are no more changes.

PR #538, opened 2019-05-01
Simulation: add a message counter in the measures

Gilthoniel:
This adds a message/packet counter alongside with the bytes counter


Gilthoniel:
Done in the scope of the "BLS over gossip protocol" student project.

PR #541, opened 2019-05-02
sets the URL when reading private.toml

ineiti:
When using app.LoadConfig to read a private.toml, this patch sets the URL correctly
to https:// if there is a WebSocketTLSCertificateKey field set.


jeffallen:
This fix is released as v3.0.9, ready to be used in Cothority's go.mod.

PR #542, opened 2019-05-03
Drop connection if an unknown error occurs

Gilthoniel:
This fixes a potential infinite loop when an unknown error occurs and
will happen every time the connection is handled. This was causing an
infinite loop in some situation.

Fixes dedis/cothority#1851


jeffallen:
Released as v3.0.10.

PR #544, opened 2019-05-08
Apply Go 1.12/Darwin file descriptor fix

jeffallen:
Also: Improve some error reporting.

### DIFF
--- a/simul/platform/platform.go
+++ b/simul/platform/platform.go
@@ -150,8 +150,10 @@ func ReadRunFile(p Platform, filename string) []*RunConfig {
 	// Decoding of the first part of the run config file
 	// where the config wont change for the whole set of the simulation's tests
 	scanner := bufio.NewScanner(file)
+	line := 0
 	for scanner.Scan() {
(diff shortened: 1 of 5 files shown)


simul/platform/platform.go
```diff
@@ -164,7 +166,7 @@ func ReadRunFile(p Platform, filename string) []*RunConfig {
 		// checking if format is good
 		vals := strings.Split(text, "=")
 		if len(vals) != 2 {
-			log.Fatal("Simulation file:", filename, " is not properly formatted ( key = value )")
+			log.Fatal("Simulation file:", filename, "line", line, "is not properly formatted ( key = value )")
```
nkcr: We miss some spaces here. I would do something like: 
`log.Fatal("Simulation file: ", filename, ":", line, " is not properly formatted ( key = value )")`

jeffallen: log.Fatal uses the same formatting rules as fmt.Println, i.e. it adds spaces for every comma.

PR #545, opened 2019-05-08
Statistics split into buckets

### LINKED ISSUE #543: Split statistics into multiple buckets
One requirement for Protean's experiments is to be able to get statistics of a given Functional Unit and for that we need to be able to get the measures from only a subset of conodes.

simulation config file:
```toml
Buckets = "a:b c:d-e:f"
```
where a-f are a valid index according to the number of hosts. The syntax is using the Go style of slice indexing and buckets are separated by a space (so we can use `strings.Fields`). Also note the usage of `-` when you need multiple ranges for a single bucket.

The `Buckets` parameter can be specified globally or per simulation if the number of hosts change.

Note: if invalid indices are used, it won't trigger any error but the buckets won't get any statistics.

Gilthoniel:
This PR adds an optional parameter to simulations to define buckets that will be filled with statistics of a subset of the hosts.

Fixes #543

### DIFF
--- a/simul/README.md
+++ b/simul/README.md
@@ -39,6 +39,21 @@ If you use the `onet.SimulationBFTree`, the following variables are also availab
 - `Depth` - the depth of the tree in levels below the root-node
 - `Rounds` - for how many rounds the simulation should run
 
+### Statistics for subset of hosts
+
+Buckets of statistics can be defined using the following variable:
+
+- `Buckets` - indices range of the buckets
+
+The parameter is a string where the buckets are separated with spaces and the ranges
+by a dash (e.g. `Buckets = "0:5 5:10-15:20"` that will create a bucket with hosts 0
+to 4 and another one with hosts 5 to 9 and 15 to 19). Range indices can be compared
+to Go slices so that the lower index is inclusive and the higher is exclusive.
+
+A file will be written per bucket and the global one containing the statistics of all
+the conodes will always be present independantly from the parameter. Each file will have
+the bucket number as suffix.
+
 ### Simulations with long setup-times and multiple measurements
 
 Per default, all rounds of an individual simulation-run will be averaged and
--- a/simul/README.md
+++ b/simul/README.md
@@ -51,7 +51,7 @@ to 4 and another one with hosts 5 to 9 and 15 to 19). Range indices can be compa
 to Go slices so that the lower index is inclusive and the higher is exclusive.
 
 A file will be written per bucket and the global one containing the statistics of all
-the conodes will always be present independantly from the parameter. Each file will have
+the conodes will always be present independently from the parameter. Each file will have
 the bucket number as suffix.
 
 ### Simulations with long setup-times and multiple measurements
--- a/simul/build.go
+++ b/simul/build.go
@@ -151,7 +151,7 @@ func RunTests(deployP platform.Platform, name string, runconfigs []*platform.Run
 
 		for j, bucketStat := range stats {
 			if j >= len(files) {
-				f, err := os.OpenFile(testFile(name, j), args, 0660)
+				f, err := os.OpenFile(generateResultFileName(name, j), args, 0660)
 				if err != nil {
 					log.Fatal("error opening test file:", err)
 				}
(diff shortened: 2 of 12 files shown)


simul/monitor/bucket_stats_test.go
```diff
@@ -0,0 +1,68 @@
+package monitor
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
```
nkcr: This is a rare occasion where true per method unit testing could've been done. 
However those two scenarios capture the essential.

simul/monitor/bucket_stats.go
```diff
@@ -0,0 +1,113 @@
...
+	for _, rule := range rr {
+		if rule.Match(host) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// BucketStats split the statistics into buckets according to network addresses
```
nkcr: splitS

simul/monitor/bucket_stats.go
```diff
@@ -0,0 +1,113 @@
...
+
+// bucketRule represents a filter that will tell if a measure must
+// be added to a bucket. It follows the logic of a slice indexing
+// so that the lower index is inclusive and the upper one is exclusive.
+type bucketRule struct {
+	low  int
+	high int
+}
+
+// newBucketRule takes a string and parse it to instantiate
```
nkcr: parseS

simul/build.go
```diff
@@ -299,8 +330,13 @@ func mkTestDir() {
 	}
 }
 
-func testFile(name string) string {
-	return "test_data/" + name + ".csv"
+func testFile(name string, index int) string {
```
nkcr: I find the name of this function misleading. Maybe `getTestFile(..)` would be more appropriate?

simul/build.go
```diff
@@ -226,10 +258,9 @@ func RunTest(deployP platform.Platform, rc *platform.RunConfig) (*monitor.Stats,
 		if err != nil {
 			return nil, err
 		}
-		return rs, nil
+		return stats, nil
 	case <-time.After(timeout):
-		monitor.Stop()
-		return rs, errors.New("Simulation timeout")
+		return nil, errors.New("Simulation timeout")
```
nkcr: error strings should not be capitalized

simul/build.go
```diff
@@ -116,25 +117,19 @@ func RunTests(deployP platform.Platform, name string, runconfigs []*platform.Run
...
-	if err != nil {
-		log.Fatal("error opening test file:", err)
-	}
+	files := []*os.File{}
 	defer func() {
-		if err := f.Close(); err != nil {
-			log.Error("Couln't close", f.Name())
+		for _, f := range files {
+			if err := f.Close(); err != nil {
+				log.Error("Couln't close", f.Name())
```
nkcr: I don't know if that applies here, but error strings should not be capitalized

simul/README.md
```diff
@@ -39,6 +39,21 @@ If you use the `onet.SimulationBFTree`, the following variables are also availab
...
+
+- `Buckets` - indices range of the buckets
+
+The parameter is a string where the buckets are separated with spaces and the ranges
+by a dash (e.g. `Buckets = "0:5 5:10-15:20"` that will create a bucket with hosts 0
+to 4 and another one with hosts 5 to 9 and 15 to 19). Range indices can be compared
+to Go slices so that the lower index is inclusive and the higher is exclusive.
+
+A file will be written per bucket and the global one containing the statistics of all
+the conodes will always be present independantly from the parameter. Each file will have
```
nkcr: independently

simul/build.go
Gilthoniel: Usually the logs are written as a correct English sentence. You're right about the errors though.

simul/build.go
Gilthoniel: Renamed as `generateResultFileName`

Gilthoniel:
@nkcr You're completely right about having comments to the test, we usually try to add those. I simply forgot so thanks for the remainder!

Comments should be fixed.

PR #547, opened 2019-05-14
Simulation: record bandwidth only before and after the simulation run

Gilthoniel:
This changes the simulation to only records the bandwidth of the servers
right before the simulation actually run to prevent messages not related
with the simulation to be taken in account.


Gilthoniel:
Either we change the current measure as it is currently done or if we are also interested in recording the bandwidth outside of the simulation, we could add another measure that specifically measures the run bandwidth.

PR #548, opened 2019-05-14
JSON over REST

kc1212:
We introduce the possible to register REST handlers using the function `RegisterRESTHandler`. It (semi) automatically computes the URL for any given callback. See the function documentation for more details. This is the first implementation so we only support GET, POST and PUT. We do not support arbitrary registration but the API can be opened up to do so.

The encoding story is not so nice. The only way I found is to write your own `MarshalJSON` and `UnmarshalJSON` for the types that use interfaces, like `kyber.Point`. But if a nested type has a json marshaller interface, then the outer type will marshal/unmarshal just fine.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -1,7 +1,6 @@
 package onet
 
 import (
-	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -11,6 +10,7 @@ import (
 	"path"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"go.dedis.ch/onet/v3/log"
@@ -207,9 +207,8 @@ func (p *ServiceProcessor) RegisterRESTHandler(f interface{}, namespace, method 
 		return err
 	}
 	var k kindGET
-	var fieldNameGET string
 	if method == "GET" {
-		k, fieldNameGET, err = prepareHandlerGET(f)
+		k, _, err = prepareHandlerGET(f)
 		if err != nil {
 			return err
 		}
@@ -223,6 +222,7 @@ func (p *ServiceProcessor) RegisterRESTHandler(f interface{}, namespace, method 
 	if err != nil {
 		return err
 	}
+	val0 := reflect.New(sh.msgType)
 
 	h := func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != method {
@@ -241,7 +241,12 @@ func (p *ServiceProcessor) RegisterRESTHandler(f interface{}, namespace, method 
 					return
 				}
 				_, num := path.Split(r.URL.EscapedPath())
-				msgBuf = []byte(fmt.Sprintf("{\"%s\": %s}", fieldNameGET, num))
+				numI64, err := strconv.Atoi(num)
+				if err != nil {
+					http.Error(w, "not a number", http.StatusBadRequest)
+					return
+				}
+				val0.Elem().Field(0).SetInt(int64(numI64))
 			case sliceGET:
 				if ok := sliceRegex.MatchString(r.URL.EscapedPath()); !ok {
 					http.Error(w, "invalid path", http.StatusNotFound)
@@ -253,8 +258,7 @@ func (p *ServiceProcessor) RegisterRESTHandler(f interface{}, namespace, method 
 					http.Error(w, err.Error(), http.StatusBadRequest)
 					return
 				}
-				encoded := base64.StdEncoding.EncodeToString(byteBuf)
-				msgBuf = []byte(fmt.Sprintf("{\"%s\": \"%s\"}", fieldNameGET, encoded))
+				val0.Elem().Field(0).SetBytes(byteBuf)
 			default:
 				http.Error(w, "invalid GET", http.StatusBadRequest)
 				return
@@ -270,17 +274,16 @@ func (p *ServiceProcessor) RegisterRESTHandler(f interface{}, namespace, method 
 				http.Error(w, err.Error(), http.StatusBadRequest)
 				return
 			}
+			if err := json.Unmarshal(msgBuf, val0.Interface()); err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
 		default:
 			http.Error(w, "unsupported method: "+r.Method, http.StatusMethodNotAllowed)
 			return
 		}
 
-		msg := reflect.New(sh.msgType).Interface()
-		if err := json.Unmarshal(msgBuf, msg); err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
-			return
-		}
-		out, tun, err := callInterfaceFunc(f, msg, false)
+		out, tun, err := callInterfaceFunc(f, val0.Interface(), false)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
--- a/processor.go
+++ b/processor.go
@@ -186,6 +186,8 @@ func prepareHandlerGET(f interface{}) (kindGET, string, error) {
 // The min/maxVersion argument represents the range of versions where the API
 // is present. If breaking changes must be made then they must use a new
 // version.
+//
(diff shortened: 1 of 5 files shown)


ineiti:
Some questions:
1. why doesn't the handler receive the request _and_ the type of request, instead of having to sign up for each request-type (GET, POST, PUT) separately?
2. why does `GET` not support `string` as input type?
3. what is possible with `json.Marshal`? Can it do `kyber.Point`, `kyber.Scalar`?
4. is it possible to replace some of the `interface{}` with proper golang-`type ... interface`?
5. is it advisable to use the protobuf-encoding to get the correct point/scalar types automatically?

kc1212:
> Some questions:
> 
>     1. why doesn't the handler receive the request _and_ the type of request, instead of having to sign up for each request-type (GET, POST, PUT) separately?
Might be possible, investigating.
> 
>     2. why does `GET` not support `string` as input type?
It's not implemented yet, but we can support it.
> 
>     3. what is possible with `json.Marshal`? Can it do `kyber.Point`, `kyber.Scalar`?
Neither `json.Marshal` and `json.Unmarshal` for interface types. You have to implement `MarshalJSON` and `UnmarshalJSON` for types that have an interface. See my PR comment and https://github.com/dedis/onet/pull/548/files#diff-683311cd2b0dbb13fea5c09795ed9d36R386 for an example. While typing this paragraph, I remembered we have this `MarshalID` thing in kyber. I will investigate and see if we can automate the JSON marshalling a bit more.  
> 
>     4. is it possible to replace some of the `interface{}` with proper golang-`type ... interface`?
I'm sticking to `interface{}` because that's how the old `RegisterHandler` works. It should be possible to use `RegisterHandler` and `RegisterRESTHandler` on the same function.
> 
>     5. is it advisable to use the protobuf-encoding to get the correct point/scalar types automatically?
I'm not sure what you mean with protobuf-encoding. Point/Scalar should be encoded using JSON, if we treat Point/Scalar as bytes, then they'll be sent around as base64 encoded strings.

ineiti:
> > 5. is it advisable to use the protobuf-encoding to get the correct point/scalar types automatically?
> I'm not sure what you mean with protobuf-encoding. Point/Scalar should be encoded using JSON, if we treat Point/Scalar as bytes, then they'll be sent around as base64 encoded strings.

In fact you answered it in 3. - using the `MarshalID`...

kc1212:
>     3. what is possible with `json.Marshal`? Can it do `kyber.Point`, `kyber.Scalar`?
Here is what I figured out: we need a custom `json.Unmarshal` to correctly figure out which implementation of `kyber.Point` to use even when the type of the point is encoded in the JSON. That's because `json.Unmarshal` takes something that has a `json.Unmarshaler` interface. You have to give `json.Unmarshal` a concrete type so that it can call `UnmarshalJSON`. This means you need to know the concrete type _before_ starting to unmarshal.

Without implementing a custom json marshaller, the easiest way is to just implement `json.Marshaler` and `json.Unmarshaler` manually. The example I provided above is quite straight forward. Basically you create another type that's exactly the same type that you want to marshal, but substitute all the `kyber.Point` and `kyber.Scalar` to `[]byte`, use `json.Unmarshal` normally on the new type, then copy all the fields back to the original type and fix the `kyber.Point` or `kyber.Scalar`.

kc1212:
>     2. why does `GET` not support `string` as input type?
Actually, doing `string` is a bit tricky due to escape characters. If the resource is usually `/x/y/thestring` and you want to read `thestring`, then u need to worry cases like `/x/y/the/string` and `/x/y/the\/string` basically we need a way to look for escape characters. I was actually using the gorilla multiplexer/router thing and it supported things like `/x/y/{s}`, then you just read the variable "s" to get your string back. But the gorilla multiplexer broke websocket upgrades so I went back to `http.ServeMux` which doesn't have that feature. To implement escape character parsing we need a lot more code (the golang `path` module doesn't do it, and I didn't see any standard libraries that do it). It'll be something like this: https://rosettacode.org/wiki/Tokenize_a_string_with_escaping

We usually get skipchain IDs or instance IDs or LTS IDs, which are all byte slices. So I'll add the string option when we have the need for it.

processor.go
```diff
@@ -143,7 +130,214 @@ func (p *ServiceProcessor) RegisterStreamingHandler(f interface{}) error {
...
+		}
+		var msgBuf []byte
+		switch r.Method {
+		case "GET":
+			if k == emptyGET {
+				msgBuf = []byte("{}")
+			} else if k == intGET {
+				ok, err := regexp.MatchString(fmt.Sprintf(`^/v\d/%s/%s/\d+$`, namespace, resource), r.URL.EscapedPath())
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusInternalServerError)
```
nkcr: 5 levels of indentation is quite a lot, would it be worth moving some part to a function?

kc1212: the high level of indentation is indeed not i deal, so I tried to refactor a bit. but I think moving the code around, either the code inside `case "GET"`, or the handler code, to another function is a bit messy because i have to pass a few arguments around. OTOH, to make the code a bit nicer, i've made the code width shorter in the latest commit.

processor.go
```diff
@@ -143,7 +130,217 @@ func (p *ServiceProcessor) RegisterStreamingHandler(f interface{}) error {
...
+					return
+				}
+				_, hexStr := path.Split(r.URL.EscapedPath())
+				byteBuf, err := hex.DecodeString(hexStr)
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusBadRequest)
+					return
+				}
+				encoded := base64.StdEncoding.EncodeToString(byteBuf)
+				msgBuf = []byte(fmt.Sprintf("{\"%s\": \"%s\"}", fieldNameGET, encoded))
```
jeffallen: I gotta say, I'm really suspicious of this "parse out the GET argument and then reformat it into JSON" technique. It seems to obscure what's really happening here which is "move the data from the GET URL to the request struct". There's already two cases, one for int and one for hex string. So in those two cases do the case-specific reflect garbage, so that once you leave the switch on method, you're ready to call callInterfaceFunc on msg directly (json.Unmarshal would only be called in POST and PUT.

processor_test.go
```diff
@@ -271,6 +286,22 @@ func newTestService(c *Context) (Service, error) {
 	if err := ts.RegisterHandler(ts.ProcessMsg); err != nil {
 		panic(err.Error())
 	}
+
+	if err := ts.RegisterRESTHandler(procRestMsgGET1, testServiceName, "GET", 3, 3); err != nil {
```
jeffallen: Those 3,3 out there on the end kind of bother me. I wonder if it would be more beautiful to have the default RegisterRESTHandler by varargs and default to 3,3 (hardcoded for now; to be changed in v4). Callers that care about versioning can opt-in by adding the min,max to the end). Dunno, this is not really important, now that I think about it.

processor_test.go
```diff
@@ -282,3 +313,307 @@ func (ts *testService) ProcessMsg(msg *testMsg) (network.Message, error) {
...
+
+type edPoint struct {
+	P kyber.Point
+}
+
+type edPointWrapper struct {
+	P []byte
+}
+
+func (p edPoint) MarshalJSON() ([]byte, error) {
```
jeffallen: So this is all very interesting, but I'm surprised to find it in *_test.go. Are you expecting to move it into Kyber later?

kc1212: `MarshalJSON` could be in kyber but having `UnmarshalJSON` in kyber is useless. suppose you want to unmarshal something, you cannot write `json.Unmarshal(buf, point)` where `point` is a `kyber.Point` because the unmarshaler doesn't know which implementation to use.

processor_test.go
kc1212: another way could be that the user can pass in a config object, 
```go
type ConfigREST struct {
    MaxVersion int
    MinVersion int
    // others if we need it, like xml support
}
```
if it's nil then we use some default

PR #552, opened 2019-05-31
Recover panics when processing client requests

Gilthoniel:
This makes sure that a handler panic will be recovered so that a
meaningful error message can be sent back instead of an empty error.

Fixes dedis/cothority#1934


processor_test.go
```diff
@@ -215,6 +215,19 @@ func TestProcessor_ProcessClientRequest(t *testing.T) {
 	}
 }
 
+// Test that the panic will be recovered and announced with crashing the server.
```
jeffallen:
```suggestion
// Test that the panic will be recovered and announced without crashing the server.
```

PR #556, opened 2019-06-12
Added AddUserInterestingGoroutine() for AfterTest

cgrigis:
Performing tests in some environments sometimes leads to leaked goroutines, which are not easy to fix without modifying an external component. Such a situation came up when integrating the Go Ethereum code into an OmniLedger contract: the Ethereum code starts goroutines to perform some caching of transactions, and it never bothers to terminate them.
As a consequence, `AfterTest()` in `log/testutil` rightfully complains about these leaking goroutines and fails the test. Sometimes one can get away with using `bct.local.Check = onet.CheckNone`, but that is not ideal.

In order to avoid polluting the "known interesting goroutines" in `log/testutil`, this PR adds a function that allows to register specific goroutines to be ignored by `AfterTest()`.

### DIFF
--- a/log/testutil.go
+++ b/log/testutil.go
@@ -31,9 +31,6 @@ func interestingGoroutines() (gs []string) {
 			strings.Contains(stack, "runtime.MHeap_Scavenger") ||
 			strings.Contains(stack, "graceful") ||
 			strings.Contains(stack, "sigqueue") ||
-			strings.Contains(stack, "stackimpact-go/internal") ||
-			// stackimpact uses persistent http client conns
-			strings.Contains(stack, "created by net/http.(*Transport).dialConn") ||
 			strings.Contains(stack, "log.MainTest") ||
 			matchesUserUninterestingGoroutine(stack) {
 			continue
--- a/log/testutil.go
+++ b/log/testutil.go
@@ -34,13 +34,40 @@ func interestingGoroutines() (gs []string) {
 			strings.Contains(stack, "stackimpact-go/internal") ||
 			// stackimpact uses persistent http client conns
 			strings.Contains(stack, "created by net/http.(*Transport).dialConn") ||
-			strings.Contains(stack, "log.MainTest") {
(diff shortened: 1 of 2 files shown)


log/testutil.go
```diff
@@ -34,15 +34,42 @@ func interestingGoroutines() (gs []string) {
...
 	return
 }
 
+var userInterestingGoroutines []string
+
+// AddUserInterestingGoroutine can be called when the environment of some
+// specific tests leaks goroutines unknown to interestingGoroutines()
+func AddUserInterestingGoroutine(newGr string) {
+	// Don't add if it's already there
+	for _, gr := range userInterestingGoroutines {
```
jeffallen: Why? There's no real harm in duplication, and anyway it's the fault of the caller, so if they want to avoid it they can.

log/testutil.go
```diff
@@ -34,15 +34,42 @@ func interestingGoroutines() (gs []string) {
...
 			continue
 		}
+
 		gs = append(gs, stack)
 	}
 	sort.Strings(gs)
 	return
 }
 
+var userInterestingGoroutines []string
```
jeffallen: They are actually UNinteresting goroutines...

log/testutil.go
```diff
@@ -34,15 +34,42 @@ func interestingGoroutines() (gs []string) {
...
+
+// AddUserInterestingGoroutine can be called when the environment of some
+// specific tests leaks goroutines unknown to interestingGoroutines()
```
jeffallen: Add: `This function is not safe for concurrent execution. The caller should add of the desired exceptions before launching any goroutines.`

log/testutil.go
cgrigis: Yup, I inadvertently inverted the meaning...

jeffallen:
Released in v3.0.15, will be referenced in cothority's go.mod in the next checkin.

PR #559, opened 2019-06-28
SendProtobufParallel

ineiti:
Problem: different methods started using different ways of sending messages in parallel
to more than one websocket-client. Sometimes these methods were even done in
parallel.

Solution: add `SendProtobufParallel` method to onet.Client with a unified interface that allows
the caller to precisely describe how to call the nodes, and reasonable defaults to let
it 'do the right thing'.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -571,6 +571,8 @@ func (c *Client) SendProtobufParallel(nodes []*network.ServerIdentity, msg inter
 				select {
 				case node = <-nodesChan:
 				default:
+				}
+				if node == nil {
 					return
 				}
 				log.Lvlf2("Asking %T from: %v - %v", msg, node.Address, node.URL)
--- a/websocket.go
+++ b/websocket.go
@@ -271,9 +271,10 @@ type destination struct {
 // Client is a struct used to communicate with a remote Service running on a
 // onet.Server. Using Send it can connect to multiple remote Servers.
 type Client struct {
-	service     string
-	connections map[destination]*websocket.Conn
-	suite       network.Suite
+	service         string
+	connections     map[destination]*websocket.Conn
+	connectionsLock map[destination]*sync.Mutex
+	suite           network.Suite
 	// if not nil, use TLS
 	TLSClientConfig *tls.Config
 	// whether to keep the connection
@@ -287,21 +288,19 @@ type Client struct {
 // connection will be started, until Close is called.
 func NewClient(suite network.Suite, s string) *Client {
 	return &Client{
-		service:     s,
-		connections: make(map[destination]*websocket.Conn),
-		suite:       suite,
+		service:         s,
+		connections:     make(map[destination]*websocket.Conn),
+		connectionsLock: make(map[destination]*sync.Mutex),
+		suite:           suite,
 	}
 }
 
 // NewClientKeep returns a Client that doesn't close the connection between
 // two messages if it's the same server.
 func NewClientKeep(suite network.Suite, s string) *Client {
-	return &Client{
-		service:     s,
-		keep:        true,
-		connections: make(map[destination]*websocket.Conn),
-		suite:       suite,
-	}
+	cl := NewClient(suite, s)
+	cl.keep = true
+	return cl
 }
 
 // Suite returns the cryptographic suite in use on this connection.
@@ -324,7 +323,9 @@ func (c *Client) newConnIfNotExist(dst *network.ServerIdentity, path string) (*w
 	// TODO we are opening a new connection for every new path?
 	// not possible to use an existing connection for the same service?
 	dest := destination{dst, path}
+	c.Lock()
 	conn, ok := c.connections[dest]
+	c.Unlock()
 
 	if !ok {
 		d := &websocket.Dialer{}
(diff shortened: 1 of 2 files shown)


websocket.go
```diff
@@ -436,6 +423,154 @@ func (c *Client) SendProtobuf(dst *network.ServerIdentity, msg interface{}, ret
...
+	// but AskNodes is 0, then AskNodes will be set to len(Roster)-StartNode.
+	//   Default: 0
+	StartNode int
+	// QuitError - if true, the first error received will be returned.
+	//   Default: false
+	QuitError bool
+	// IgnoreNodes is a set of nodes that will not be contacted. They are counted towards
+	// AskNodes and StartNode, but not contacted.
+	//   Default: false
+	IgnoreNodes []*network.ServerIdentity
```
kc1212: hmm.. I wonder if it'll be easier to just use a list of server identities instead of a roster in `SendProtobufParallel`. we give the caller the responsibility of creating a list of nodes that it wants to use in any way he wants so that the caller is not forced to use the `IgnoreNodes` convention in `ParallelOptions`. but the other options (int and bool ones) in `ParallelOptions` are nice to keep.

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/montanaflynn/stats v0.5.0
 	github.com/stretchr/testify v1.3.0
-	go.dedis.ch/kyber/v3 v3.0.3
+	go.dedis.ch/kyber/v3 v3.0.2
```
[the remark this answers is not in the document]
ineiti: Don't know what happened.

websocket.go
ineiti: I like the `IgnoreNodes` because it's a simple fix for most of the `GetBlock` calls in byzcoin. I used it in the skipchain-api to exclude the calling node. Example how this looks for the client:
https://github.com/dedis/cothority/pull/1984/files#diff-dde946dfedf27eeee85bb4b483dc2687R998
and the api call:
https://github.com/dedis/cothority/pull/1984/files#diff-55e058d9e0138dcc35f85d083a985a9bR256

ineiti:
> I like the proposal but I'm not 100% convinced that it works because c.Send has a mutex inside.

Nice catch - changed the mutex in there to only protect the common structures.

websocket.go
```diff
@@ -387,29 +388,34 @@ func (c *Client) newConnIfNotExist(dst *network.ServerIdentity, path string) (*w
 // Send will marshal the message into a ClientRequest message and send it.
 func (c *Client) Send(dst *network.ServerIdentity, path string, buf []byte) ([]byte, error) {
 	c.Lock()
-	defer c.Unlock()
-
 	conn, err := c.newConnIfNotExist(dst, path)
 	if err != nil {
 		return nil, err
```
kc1212: missing unlock if connection fails

websocket_test.go
```diff
@@ -667,6 +667,70 @@ func TestWebSocket_Streaming_Parallel(t *testing.T) {
...
+	count, list = po.GetList(nodes)
+	require.Equal(t, 1, count)
+	require.Equal(t, 1, len(list))
+
+	po.StartNode = 1
+	count, list = po.GetList(nodes)
+	require.Equal(t, 1, count)
+	require.Equal(t, 1, len(list))
+}
+
```
kc1212: could you add a test that runs `SendProtobufParallel`? just to make sure it continues to work in the future

ineiti:
OK, made `Client.Send` somewhat parallel running, at least as far as it makes sense with websockets: different connections can run in parallel, but if you want to send a message through an existing connection (`Client.keep == true`, or message in transit), then it waits for the previous message to be done.

websocket.go
```diff
@@ -318,7 +319,9 @@ func (c *Client) newConnIfNotExist(dst *network.ServerIdentity, path string) (*w
 	// TODO we are opening a new connection for every new path?
```
Gilthoniel: Not related to the PR but this TODO can go away imho. Connections are tight to their end points and it's not designed to be reused.

websocket.go
```diff
@@ -436,6 +456,155 @@ func (c *Client) SendProtobuf(dst *network.ServerIdentity, msg interface{}, ret
...
+				if node == nil {
+					return
+				}
+				log.Lvlf2("Asking %T from: %v - %v", msg, node.Address, node.URL)
+				reply, err := c.Send(node, path, buf)
+				if err != nil {
+					log.Lvl2("Error while sending to node:", node, err)
+					errChan <- err
+				} else {
+					finish.Do(func() { close(nodesChan) })
```
Gilthoniel: If all the requests fail, the go routines will be stuck trying to read the next node (as there is no node left)

ineiti: Are you referring to the lines
```
case node = <-nodesChan:
default:
```
?
Then it's OK, because the `default` is the fall-through option. I can put the `return` in the default. But it seemed to me that a `nil` node can appear in the channel... Perhaps not.

ineiti: Now I know why the `if node == nil` came after the select:
- if there is a value, `node` is populated
- if the channel is empty, it goes in `default`, and `node == nil`
- if the channel is closed, `node` is populated with `nil` !?!

PR #561, opened 2019-07-09
Return if successful

ineiti:
`SendProtobufParallel` didn't return when a successful `Send` happened.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -584,6 +584,7 @@ func (c *Client) SendProtobufParallel(nodes []*network.ServerIdentity, msg inter
 					finish.Do(func() { close(nodesChan) })
 					siChan <- node
 					replyChan <- reply
+					return
 				}
 			}
 		}()


websocket.go
```diff
@@ -584,6 +584,7 @@ func (c *Client) SendProtobufParallel(nodes []*network.ServerIdentity, msg inter
 					finish.Do(func() { close(nodesChan) })
 					siChan <- node
 					replyChan <- reply
+					return
```
kc1212: just to check my understanding: the go-routine will return anyway if we succeed because `nodesChan` is closed so we will return on line 576, is that right?

ineiti: that's what I thought, but for some reason it doesn't :( At least as seen by my tests...

ineiti: I'll try it again, you're right, the `return` looks superfluous!

ineiti:
Good thinking, but not done... See #562

PR #563, opened 2019-07-09
All error path

ineiti:
The all-error path was wrong and not correctly handled. Also update the test to check if it behaves correctly now.

### DIFF
--- a/websocket_test.go
+++ b/websocket_test.go
@@ -9,6 +9,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"errors"
 	"io/ioutil"
 	"math/big"
 	"net"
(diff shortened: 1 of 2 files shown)


websocket_test.go
```diff
@@ -765,17 +779,36 @@ const serviceWebSocket = "WebSocket"
 
 type ServiceWebSocket struct {
 	*ServiceProcessor
+	Errors int
```
kc1212: do you need a lock for accessing this field?

ineiti: No, because golang doesn't complain.

I meant: no, because it's only accessed once per call, and each service has its own `Errors`. Also, the `SendProtobufParallel` call should be done when the counting starts. So a golang race-error would be an indication of the call not having finished.

PR #565, opened 2019-07-22
Racy websocket closing

ineiti:
- Fix a race in the closing of the websocket
- Stop contacting other nodes in SendProtobufParallel if one node replied successfully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ include $(shell go env GOPATH)/src/github.com/dedis/Coding/bin/Makefile.base
 # to `make test_playground`.
 test_playground:
 	for a in $$( seq 100 ); do \
-	  if DEBUG_TIME=true go test -v -race -short -run TestClient_SendProtobufParallel > log.txt 2>&1; then \
+	  if DEBUG_TIME=true go test -v -race -short > log.txt 2>&1; then \
 		  echo Successfully ran \#$$a at $$(date); \
 		else \
 		  echo Failed at $$(date); \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -7,8 +7,8 @@ include $(shell go env GOPATH)/src/github.com/dedis/Coding/bin/Makefile.base
 # to `make test_playground`.
 test_playground:
 	for a in $$( seq 100 ); do \
-	  if DEBUG_TIME=true go test -v -race -short > log.txt 2>&1; then \
-		  echo Successfully ran at $$(date); \
+	  if DEBUG_TIME=true go test -v -race -short -run TestClient_SendProtobufParallel > log.txt 2>&1; then \
+		  echo Successfully ran \#$$a at $$(date); \
```
kc1212: please revert this part

websocket.go
```diff
@@ -319,15 +319,29 @@ func (c *Client) closeSingleUseConn(dst *network.ServerIdentity, path string) {
...
 
+	// c.Lock protects the connections and connectionsLock map
+	// c.connectionsLock is held as long as the connection is in use - to avoid that two
+	// processes send data over the same websocket concurrently.
 	dest := destination{dst, path}
 	c.Lock()
-	conn, ok := c.connections[dest]
+	connLock, exists := c.connectionsLock[dest]
+	if !exists {
+		c.connectionsLock[dest] = &sync.Mutex{}
```
kc1212: do we need to delete the entry from the map when the send is done?

Makefile
ineiti: Done - although it's only for the `test_playground` target, which is only used when debug-testing with Travis. I kept the change of `\#$$a`, because this was missing...

websocket.go
ineiti: The current locking needs the `connectionLock` to stay the same when a caller uses the `Client` for the same node/path destination twice in a row. The connection will be closed between the two calls and deleted in `c.connections`, but the lock must stay the same.

PR #569, opened 2019-08-08
Use a custom decoder to refuse a message

Gilthoniel:
This adds a custom decoder when sending in parallel so that messages
can be analyzed and refused (e.g. old block index)

Related to dedis/cothority#2002

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -615,7 +615,7 @@ func (c *Client) SendProtobufParallelWithDecoder(nodes []*network.ServerIdentity
 				err := decoder(reply, ret)
 				if err != nil {
 					errs = append(errs, err)
-					// Reply avorted so we empty the last entry of this channel.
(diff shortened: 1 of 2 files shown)


websocket.go
```diff
@@ -612,9 +612,19 @@ func (c *Client) SendProtobufParallel(nodes []*network.ServerIdentity, msg inter
 		select {
 		case reply := <-replyChan:
 			if ret != nil {
-				return <-siChan, protobuf.DecodeWithConstructors(reply, ret,
-					network.DefaultConstructors(c.suite))
+				err := decoder(reply, ret)
+				if err != nil {
+					errs = append(errs, err)
+					// Reply avorted so we empty the last entry of this channel.
```
kc1212:
```suggestion
					// Reply aborted so we empty the last entry of this channel.
```

PR #570, opened 2019-08-09
Fix goroutine leak with SendProtobufParellel

Gilthoniel:
Channels must be buffered to a high enough value for scenarios where
multiple conodes send a correct reply, as closing will stuck the reply
channel and thus having one or more leaking goroutines.

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -577,10 +577,9 @@ func (c *Client) SendProtobufParallelWithDecoder(nodes []*network.ServerIdentity
 	parallel, nodesChan := opt.GetList(nodes)
 	nodesNbr := len(nodesChan)
 	errChan := make(chan error, nodesNbr)
-	replyChan := make(chan []byte, parallel)
-	siChan := make(chan *network.ServerIdentity, parallel)
+	replyChan := make(chan []byte, nodesNbr)
+	siChan := make(chan *network.ServerIdentity, nodesNbr)
 	closed := make(chan bool)
-	finish := &sync.Once{}
 	for g := 0; g < parallel; g++ {
 		go func() {
 			for {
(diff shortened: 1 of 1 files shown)


websocket.go
```diff
@@ -621,18 +620,19 @@ func (c *Client) SendProtobufParallelWithDecoder(nodes []*network.ServerIdentity
...
 		case err := <-errChan:
 			if opt.Quit() {
+				close(closed)
 				return nil, err
 			}
 			errs = append(errs, err)
 		}
 	}
+
+	close(closed)
```
ineiti: I don't understand how you can close a closed channel again without raising a panic - is this code-path tested?

Gilthoniel: there is a close after each return ...

ineiti: Yes, and I put it in a `sync.Once` because it gave me panics otherwise. So I'm surprised that you can close the channel more than once...

ineiti: Can you have a look if the following commit might have introduced the error you're trying to fix here?

https://github.com/dedis/onet/commit/c575e990469fa077aa77a406612f8ed694039bf3#diff-8990d2c9c7f1690e60641d425c8932f7L600

It seems to me that it's more correct to close the channel in the go-routine than in the consumer for-loop, as the consumer might have to wait to be able to close the channel. As soon as a correct message is received, the channel should be closed. Then you won't have problems with the channel overflowing, and you can keep it with `parallel` length.

ineiti:
OK, now I think I get the problem with the current state of `master`: the decoding of the received message is done outside of the producer-go-routine, which makes it possible to fill up the `replyChan`.

I created a cleaned-up version of this in #572 - please tell me if this is a good approach.

jeffallen:
Used solution in #572.

PR #572, opened 2019-08-27
Check decoding in go-routine

ineiti:
To avoid too many requests on nodes, decode the correct replies already in the
go-routine. This avoids the channel filling up with potential correct
data that will only be checked in the consumer-loop.

This also simplifies the number of channels used.

Replaces #570

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -558,7 +558,7 @@ func (po *ParallelOptions) Quit() bool {
 	return po.QuitError
 }
 
-// Decoder is function that takes the data and the interface to fill in
+// Decoder is a function that takes the data and the interface to fill in
 // as input and decodes the message.
 type Decoder func(data []byte, ret interface{}) error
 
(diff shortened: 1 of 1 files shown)


ineiti:
> Should't the first `select case` priorize the `case <- done` statement? With something like:

As far as I understand go, there is no such thing as _priorities_ in a `select` statement - they're chosen at random, see here:

https://yourbasic.org/golang/select-explained/

nkcr:
> As far as I understand go, there is no such thing as _priorities_ in a `select` statement - they're chosen at random

Yes, exactly. Except for the default case, which is executed only when no other cases can be run. This is why in my example, for the first select, I kept only the `case <-done:` and moved the rest in the default case. This ensures that we always check for the `done` Chanel.

ineiti:
OK - got it. Thanks for explaining ;) I also put the `select` in a separate variable, else the indention gets a bit out of control...

ineiti:
And btw: I checked on play.golang.org, and a closed channel does not have priority over other channels that have data available. It is random also including closed channels.

nkcr:
Do you want to wait on Galylor's review, or should I merge it?

ineiti:
It was an answer to #570, so if you know how to reproduce the error that got @kc1212 to write #570, then please look if this PR fixes it. I prefer this over #570, so if you go ahead and merge it, you should also close #570.

PR #576, opened 2019-09-23
Reload ssl certificate

### LINKED ISSUE #573: need to reload LE certificate
When onet is using an externally-refreshed certificate (like when `WebSocketTLSCertificate` is set in private.toml in a conode) and the certificate is refreshed by a cronjob on the server, onet does not detect this and reload the certificate.

One solution is to just add a post-refresh job that restarts the conode. But it would be better if onet monitored the expiration date in the certificate it is using, and starts trying reloads 24 hours before.

Gilthoniel:
Fixes #573

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -143,22 +143,40 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
 
 	// Set Websocket TLS if possible
 	if hc.WebSocketTLSCertificate != "" && hc.WebSocketTLSCertificateKey != "" {
-		tlsCertificate, err := hc.WebSocketTLSCertificate.Content()
-		if err != nil {
-			return nil, nil, fmt.Errorf("getting WebSocketTLSCertificate content: %v", err)
-		}
-		tlsCertificateKey, err := hc.WebSocketTLSCertificateKey.Content()
-		if err != nil {
-			return nil, nil, fmt.Errorf("getting WebSocketTLSCertificateKey content: %v", err)
-		}
-		cert, err := tls.X509KeyPair(tlsCertificate, tlsCertificateKey)
-		if err != nil {
-			return nil, nil, fmt.Errorf("loading X509KeyPair: %v", err)
-		}
+		if hc.WebSocketTLSCertificate.CertificateURLType() == String {
+			tlsCertificate, err := hc.WebSocketTLSCertificate.Content()
+			if err != nil {
+				return nil, nil, fmt.Errorf("getting WebSocketTLSCertificate content: %v", err)
+			}
+			tlsCertificateKey, err := hc.WebSocketTLSCertificateKey.Content()
+			if err != nil {
+				return nil, nil, fmt.Errorf("getting WebSocketTLSCertificateKey content: %v", err)
+			}
+			cert, err := tls.X509KeyPair(tlsCertificate, tlsCertificateKey)
+			if err != nil {
+				return nil, nil, fmt.Errorf("loading X509KeyPair: %v", err)
+			}
 
-		server.WebSocket.Lock()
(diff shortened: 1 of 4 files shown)


app/config.go
```diff
@@ -143,22 +143,44 @@ func ParseCothority(file string) (*CothorityConfig, *onet.Server, error) {
...
-			return nil, nil, fmt.Errorf("getting WebSocketTLSCertificateKey content: %v", err)
-		}
-		cert, err := tls.X509KeyPair(tlsCertificate, tlsCertificateKey)
-		if err != nil {
-			return nil, nil, fmt.Errorf("loading X509KeyPair: %v", err)
-		}
+		if hc.WebSocketTLSCertificate.CertificateURLType() == File &&
+			hc.WebSocketTLSCertificateKey.CertificateURLType() == File {
+			// Use the reloader only when both are files as it doesn't
+			// make for string embedded certififcates.
```
ineiti:
```suggestion
			// make sense for string embedded certififcates.
```

websocket.go
```diff
@@ -21,6 +22,72 @@ import (
...
+// CertificateReloader takes care of reloading a TLS certificate when
+// requested.
+type CertificateReloader struct {
+	sync.RWMutex
+	cert     *tls.Certificate
+	certPath string
+	keyPath  string
+}
+
+// NewCertificateReloader takes two file paths as parameter that shall contain
```
ineiti:
```suggestion
// NewCertificateReloader takes two file paths as parameter that contain
```

websocket.go
```diff
@@ -21,6 +22,72 @@ import (
...
+		// Here we know the leaf has been parsed successfully as an error
+		// would have been thrown otherwise.
+		if cr.cert == nil || exp.After(cr.cert.Leaf.NotAfter) {
+			// Certificate has expired so we try to load the new one.
+
+			// Free the read lock to be able to reload.
+			cr.RUnlock()
+			err := cr.reload()
+			if err != nil {
+				return nil, err
```
jeffallen: This return, plus the pending defer, will trigger a duplicate call to RUnlock(), which might be ok, I don't know.

I think mixing "defer unlock" and "need to just unlock this before this call and relock it after" is asking for trouble, I'm still looking and thinking if there's another way.

websocket.go
```diff
@@ -21,6 +22,72 @@ import (
...
+	return loader, loader.reload()
+}
+
+func (cr *CertificateReloader) reload() error {
+	newCert, err := tls.LoadX509KeyPair(cr.certPath, cr.keyPath)
+	if err != nil {
+		return err
+	}
+
+	cr.Lock()
```
jeffallen: What if the pre-condition to calling reload() is that you hold the read lock, and then reload() takes responsibility for releasing the read lock and getting the write lock if it needs to? That means one more lock/unlock in NewCertificateReloader but then simplified code in GetCertificateFunc().

The certificate function is called 1 time for every incoming TLS connection, so it really needs to be as lightweight as possible.

websocket.go
```diff
@@ -21,6 +22,72 @@ import (
 	graceful "gopkg.in/tylerb/graceful.v1"
 )
 
+const certificateReloaderLeeway = 1 * time.Hour
+
+// CertificateReloader takes care of reloading a TLS certificate when
+// requested.
+type CertificateReloader struct {
+	sync.RWMutex
```
jeffallen: Nice, this is a very rare case where RWMutex makes sense. 👍

websocket.go
Gilthoniel: The defer is declared after the if clause. It could actually be replaced by a copy of cert then unlock then return but it won't have a duplicate unlock.

PR #580, opened 2019-10-07
Add pprof to onet

jeffallen:
Make the pprof HTTP interface available on the webserver, if and
only if the server admin opts in by setting the environment variable
ONET_ALLOW_PPROF.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,3 +22,5 @@ require (
 	gopkg.in/tylerb/graceful.v1 v1.2.15
 	rsc.io/goversion v1.2.0
 )
+
+go 1.13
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -22,3 +22,5 @@ require (
 	gopkg.in/tylerb/graceful.v1 v1.2.15
 	rsc.io/goversion v1.2.0
 )
+
+go 1.13
```
Gilthoniel: Does that mean you need go >= v1.13 ? That would break our policy to support one version below if yes.

jeffallen: The go 1.13 tools add that automatically, but go 1.12.10 can still build onet, so I guess it is ok.

PR #586, opened 2019-10-14
V4 initial commit

Gilthoniel:


ineiti:
Is there a short description on how you would like to go on from here? What is allowed? What should be the workflow? How can we break APIs? When will v4 be released?

PR #587, opened 2019-10-15
Implementation of the cipher suite registry

### LINKED ISSUE #579: Make Onet independent from Kyber
Those are the different points to change for Onet to be independent:
[] ServerToml is now tightly bound to kyber because of the public identity of a server that is a kyber point.
[]

Gilthoniel:
- Add a _ciphersuite_ package that contains the abstraction of the cryptographic primitives.
- Add a builder.go file that contains the implementation of a server builder.
- Remove the public API to create a server.
- Remove the service factory to have a non-static service registration through the builder.

Closes #579


ineiti:
Is there a document describing what you're doing here?

Gilthoniel:
Question:
- Rename `ciphersuite` to something shorter ? `csuite` ?

Gilthoniel:
Follow-ups:
-

README.md
```diff
@@ -68,6 +68,70 @@ should be no API breaking changes.
...
+Each service can have a different cipher suite associated with it. The advantage
+is that it will generate a different key pair for this service so it won't share
+the master key pair. That means the cipher suite can be of the same type as the
+master one if you only want to generate a different key pair for the service.
+
+Onet only defines the primitives that it needs to operate but you are free when
+creating a service to extend the base interface and force the registration of the
+service with this extended cipher suite:
+
+```go
```
ineiti: I would prefer seeing the actual cipher suite definition with some explanation here, and then how to extend it...

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+type Name = string
+
+// Nameable binds a structure to a cipher.
+type Nameable interface {
+	Name() Name
+}
+
+// CipherData is a self-contained message type that can be used
+// over the network in the contrary of the interfaces.
+type CipherData struct {
```
ineiti: Why don't you use the same system as with the kyber suite binaryMarshaller? It reminds me of https://xkcd.com/927/

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+	Data []byte
+	Name Name
+}
+
+func (d *CipherData) String() string {
+	buf := append([]byte(d.Name), d.Data...)
```
ineiti: For a string I would do something like
```
fmt.Sprintf("%s: %x", d.Name, d.Data)
```

makes it a _little bit_ nicer...

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+
+	size, err = w.Write(d.Data)
+	n += int64(size)
+	return
+}
+
+// MarshalText implements the encoding interface TextMarshaler so that
+// it can be serialized in format such as TOML.
+func (d *CipherData) MarshalText() ([]byte, error) {
+	name := []byte(d.Name)
```
ineiti: If it ends up in a `toml` file it will be read by persons, so I'd prefer to have a clear-text representation of the `Name`. So I would rather use `String()` here.
Also, I don't see why you want to add the size here?

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	d.Name = string(buf[sizeLength : sizeLength+size])
+	d.Data = buf[sizeLength+size:]
+	return nil
+}
+
+// Packable provides the primitives necessary to make network
+// messages out of interfaces.
+type Packable interface {
```
ineiti: How does this work together with the current way of registering the interfaces to the `protobuf` package?

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Unpack takes a cipher data container and recreate the corresponding
+	// element. If the type does not match or an decoding error occurs, it
+	// should return the error.
+	Unpack(data *CipherData) error
+}
+
+// PublicKey represents one of the two sides of an asymmetric key pair
+// which can be safely shared publicly.
+type PublicKey interface {
```
ineiti: Please make sure that this makes sense from a crypto point of view. I'm not sure whether you want to have public/private, point/scalar, xxx/yyy here. Also, I don't know how it works in the case of lattice-based crypto.

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+	fmt.Stringer
+}
+
+// CipherSuite provides the primitive needed to create and verify
+// signatures using an asymmetric key pair.
+type CipherSuite interface {
+	Nameable
+
+	// PublicKey must return an implementation of a public key.
+	PublicKey() PublicKey
```
ineiti: Just to make sure: did you discuss this with Christian Mouchet, Philipp Jovanovic and Bryan? You're creating a new very basic interface to cryptographic primitives. So you really want to get this right.

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// SecretKey must return an implementation of a secret key.
+	SecretKey() SecretKey
```
ineiti: E.g.: the SecretKey here does not have any functionality, how can you get from this SecretKey to a key where you can do arithmetic operations?

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Signature must return an implementation of a signature.
+	Signature() Signature
+
+	// KeyPair must return a random secret key and its associated public key.
+	KeyPair() (PublicKey, SecretKey)
+
+	// Sign must produce a signature that can be validated by the
+	// associated public key of the secret key.
+	Sign(sk SecretKey, msg []byte) (Signature, error)
```
ineiti: Not sure: is this enough for lattice-based crypto? Should there be a more generic interface?

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Verify must return nil when the signature is valid for the
+	// message and the public key. Otherwise it should return the
+	// reason of the invalidity.
+	Verify(pk PublicKey, signature Signature, msg []byte) error
```
ineiti: Same thing here: are there more parameters needed in the case of lattigo?

ciphersuite/unsecure_csuite.go
```diff
@@ -0,0 +1,180 @@
...
+	"bytes"
+	"crypto/rand"
+	"encoding/hex"
+
+	"golang.org/x/xerrors"
+)
+
+const nonceLength = 16
+
+// UnsecureCipherSuiteName is the reference name for the cipher suite
```
ineiti: If this is only used for testing, then why isn't it in `_test.go`?

ineiti: Why is it unsecure? Why do you need an unsecure suite?

ciphersuite/unsecure_csuite.go
```diff
@@ -0,0 +1,180 @@
...
+// KeyPair generates a valid pair of public and secret key for the
+// unsecure cipher suite.
+func (c *UnsecureCipherSuite) KeyPair() (PublicKey, SecretKey) {
+	return newUnsecurePublicKey(), newUnsecureSecretKey()
+}
+
+// Sign takes a secret key and a message and returns the signature of
+// the message that can be verified by the associated public key.
+func (c *UnsecureCipherSuite) Sign(sk SecretKey, msg []byte) (Signature, error) {
+	if len(msg) == 0 {
```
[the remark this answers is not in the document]
ineiti: I would at least use `sha256` here in the 'signature'. Then you have a constant length signature, which is much more accurate, even for a `UnsecureCipherSuite`.

simulation.go
```diff
@@ -250,12 +244,13 @@ type SimulationBFTree struct {
 // 'addresses'. The network.Address(es) created are of type TLS or PlainTCP,
 // depending on the value 'TLS' in 'sc'.
 func (s *SimulationBFTree) CreateRoster(sc *SimulationConfig, addresses []string, port int) {
+	if sc.Builder == nil {
+		// TODO: must have a better way to inforce to get the builder.
```
tharvik: `SimulationConfig` and `SimulationBFTree` are quite close, maybe creating the latter with help of the first (thus ensuring to have a builder)?

tharvik: They even both have a `TLS` key, with the same meaning.

tree.go
```diff
@@ -417,68 +389,45 @@ var RosterTypeID = network.RegisterMessage(Roster{})
...
-		}
-
-		for _, srvid := range id.ServiceIdentities {
-			_, err = srvid.Public.MarshalTo(h)
-			if err != nil {
-				log.Error(err)
-			}
-		}
-	}
+	// Take a copy of ids, in case the caller tries to change it later.
```
tharvik: Just as he can via `Roster.List` afterward.

tree.go
```diff
@@ -384,11 +359,8 @@ func (tm *TreeMarshal) MakeTreeFromList(parent *TreeNode, ro *Roster) (*TreeNode
 // A Roster is a list of ServerIdentity we choose to run some tree on it ( and
 // therefor some protocols). Access is not safe from multiple goroutines.
 type Roster struct {
-	// Deprecated: ID can be generated using the list of server identities. Use GetID instead.
-	ID RosterID
 	// List is the list of actual entities.
-	List      []*network.ServerIdentity
-	Aggregate kyber.Point
+	List []*network.ServerIdentity
```
tharvik: `type Roster []*network.ServerIdentity`

tree.go
```diff
@@ -417,68 +389,45 @@ var RosterTypeID = network.RegisterMessage(Roster{})
...
-			if e.Public != nil {
-				agg = agg.Add(agg, e.Public)
-			}
-		}
-		r.Aggregate = agg
-	}
-	return r
+// Len returns the length of the roster.
+func (ro *Roster) Len() int {
+	return len(ro.List)
```
tharvik: Why? It's shorter to write `len()` and avoid the call.

tree.go
```diff
@@ -417,68 +389,45 @@ var RosterTypeID = network.RegisterMessage(Roster{})
...
-func (ro *Roster) GetID() (RosterID, error) {
+func (ro *Roster) GetID() RosterID {
 	h := sha256.New()
 	for _, id := range ro.List {
-		_, err := id.Public.MarshalTo(h)
+		_, err := id.PublicKey.WriteTo(h)
 		if err != nil {
-			return RosterID{}, xerrors.Errorf("marshaling: %v", err)
+			// An error at that point would mean something worst is
+			// happening, so we keep the function simple and panic.
```
tharvik: Or, we can have a `RosterBuilder` which populate the `ID` field when building :P

treenode.go
```diff
@@ -656,47 +650,74 @@ func (n *TreeNodeInstance) OnDoneCallback(fn func() bool) {
...
-func (n *TreeNodeInstance) Public() kyber.Point {
-	serviceName := ServiceFactory.Name(n.token.ServiceID)
+func (n *TreeNodeInstance) PublicKey() ciphersuite.PublicKey {
+	_, name := n.overlay.server.serviceManager.serviceByID(n.token.ServiceID)
 
-	return n.Host().ServerIdentity.ServicePublic(serviceName)
+	data := n.Host().ServerIdentity.ServicePublic(name)
+	pk, err := n.overlay.cr.UnpackPublicKey(data)
+	if err != nil {
+		log.Error(err)
```
tharvik: `log.Panicf`

treenode.go
```diff
@@ -656,47 +650,74 @@ func (n *TreeNodeInstance) OnDoneCallback(fn func() bool) {
...
+func (n *TreeNodeInstance) NodePublic(si *network.ServerIdentity) ciphersuite.PublicKey {
+	_, name := n.overlay.server.serviceManager.serviceByID(n.token.ServiceID)
 
-	return si.ServicePublic(serviceName)
+	data := si.ServicePublic(name)
+	pk, err := n.overlay.cr.UnpackPublicKey(data)
+	if err != nil {
+		log.Error(err)
+		panic("Couldn't unpack the public key of the distant node. Please check " +
+			"the of the server.")
```
tharvik: "Sure, I'll check the of the server."

ciphersuite/registry.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return sig, nil
+}
+
+// Verify takes a public key, a signature and a message and performs a verification
+// that will return an error if the signature does not match the message. It
+// will also return an error if the cipher suite is unknown.
+func (cr *Registry) Verify(pk PublicKey, sig Signature, msg []byte) error {
+	if pk.Name() != sig.Name() {
+		return xerrors.New("mismatching cipher names")
```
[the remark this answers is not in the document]
ineiti: I think cothority still wants to keep backward-compatibility with 1.12, which has no `xerrors` in the stdlib yet.

Gilthoniel:
I heard your thoughts! Thanks, I'm going to create multiple small PRs with independent modules and a last one to make them work together.

PR #592, opened 2019-10-28
Cipher suite interface to replace Kyber

Gilthoniel:
This adds a cipher suite interface that will be used instead of
Kyber abstractions to allow any kind of asymmetric cryptography.

It also implements the interface by using the Ed25519 scheme
provided by the language.

Related to #579

### DIFF
--- a/ciphersuite/ciphersuite.go
+++ b/ciphersuite/ciphersuite.go
@@ -1,0 +1,285 @@
+// Package ciphersuite defines the interfaces that Onet needs to setup
+// a secure channel between the conodes. It is built around a cipher suite
+// interface that provides the cryptographic primitives.
+//
+// The package also provides a cipher suite implementation that is using the
+// Ed25519 signature scheme.
+//
+// As a server could use multiple cipher suites, the package implements a
+// cipher registry that takes an implementation of a cipher suite and
+// registered using the name of the suite.
+//
+// Public keys and signatures may need to be transmitted over the network and
+// interfaces cannot be used as is. That is why every the different elements
+// can be packed as CipherData. The registry provides functions to unpack
+// them as the structure is self-contained.
+package ciphersuite
+
+import (
+	"bytes"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"io"
+
+	"golang.org/x/xerrors"
+)
+
+var sizeLength = 32 / 8
+
+// Name is the type that can differentiate multiple ciphers.
+type Name = string
+
+// Nameable binds a structure to a cipher.
+type Nameable interface {
+	Name() Name
+}
+
+// CipherData is a self-contained message type that can be used
+// over the network in the contrary of the interfaces.
+type CipherData struct {
+	Data       []byte
+	CipherName Name
+}
+
+// Name returns the name of the cipher suite compatible with the data
+// contained in the raw structure.
+func (d *CipherData) Name() Name {
+	return d.CipherName
+}
+
+func (d *CipherData) String() string {
+	buf := append([]byte(d.Name()), d.Data...)
+	return hex.EncodeToString(buf)
+}
+
+// Equal verifies if both self and other are deeply equal.
+func (d *CipherData) Equal(other *CipherData) bool {
+	return d.Name() == other.Name() && bytes.Equal(d.Data, other.Data)
+}
+
+// Clone returns a clone of the cipher data.
+func (d *CipherData) Clone() *CipherData {
+	data := make([]byte, len(d.Data))
+	copy(data, d.Data)
+	return &CipherData{
+		CipherName: d.Name(),
+		Data:       data,
+	}
+}
+
+// WriteTo implements the io.WriteTo interface so that the cipher
+// data can be written into any standard writer (e.g. hash).
+func (d *CipherData) WriteTo(w io.Writer) (n int64, err error) {
+	var size int
+	size, err = w.Write([]byte(d.Name()))
+	n += int64(size)
+	if err != nil {
+		return
+	}
+
+	size, err = w.Write(d.Data)
+	n += int64(size)
+	return
+}
+
+// MarshalText implements the encoding interface TextMarshaler so that
+// it can be serialized in format such as TOML.
+func (d *CipherData) MarshalText() ([]byte, error) {
+	name := []byte(d.Name())
+	size := make([]byte, sizeLength)
+	binary.LittleEndian.PutUint32(size, uint32(len(name)))
+
+	// Buffer starts with the size of the cipher suite name, then the name
+	// and finally the data.
+	data := append(append(size, name...), d.Data...)
+
+	buf := make([]byte, hex.EncodedLen(len(data)))
+	hex.Encode(buf, data)
+	return buf, nil
+}
+
+// UnmarshalText implements the encoding interface TextUnmarshaler so that
+// format such as TOML can deserialize the data.
+func (d *CipherData) UnmarshalText(text []byte) error {
+	buf := make([]byte, hex.DecodedLen(len(text)))
+	_, err := hex.Decode(buf, text)
+	if err != nil {
+		return xerrors.Errorf("decoding hex: %v", err)
+	}
+
+	if len(buf) < sizeLength {
+		return xerrors.Errorf("data is too small")
+	}
+
+	size := int(binary.LittleEndian.Uint32(buf[:sizeLength]))
+	if len(buf) < sizeLength+size {
+		return xerrors.Errorf("data is too small")
+	}
+
+	d.CipherName = string(buf[sizeLength : sizeLength+size])
+	d.Data = buf[sizeLength+size:]
+	return nil
+}
+
+// RawPublicKey is a raw data structure of a public key implementation.
+type RawPublicKey struct {
+	*CipherData
+}
+
+// NewRawPublicKey returns an instance of a public key.
+func NewRawPublicKey(name Name, data []byte) *RawPublicKey {
+	return &RawPublicKey{
+		CipherData: &CipherData{
+			Data:       data,
+			CipherName: name,
+		},
+	}
+}
+
+// Raw returns the raw data of a public key. It is implemented to allow
+// a raw public key to be compatible with the interface.
+func (raw *RawPublicKey) Raw() *RawPublicKey {
+	return raw
+}
+
+// Equal returns true when the two data structure contains the same public
+// key.
+func (raw *RawPublicKey) Equal(other PublicKey) bool {
+	data := other.Raw()
+	return data.CipherData.Equal(raw.CipherData)
+}
+
+// Clone returns a clone of the raw public key.
+func (raw *RawPublicKey) Clone() *RawPublicKey {
+	return &RawPublicKey{CipherData: raw.CipherData.Clone()}
+}
+
+// UnmarshalText converts the raw public key back from a text marshaling.
+func (raw *RawPublicKey) UnmarshalText(text []byte) error {
+	raw.CipherData = &CipherData{}
+	return raw.CipherData.UnmarshalText(text)
+}
+
+// RawSecretKey is a raw data structure of a secret key implementation.
+type RawSecretKey struct {
+	*CipherData
+}
+
+// NewRawSecretKey returns an instance of a raw secret key.
+func NewRawSecretKey(name Name, data []byte) *RawSecretKey {
+	return &RawSecretKey{
+		CipherData: &CipherData{
+			CipherName: name,
+			Data:       data,
+		},
+	}
+}
+
+// Raw returns the raw data of a secret key. It is implemented to allow
+// a raw secret key to be compatible with the interface.
+func (raw *RawSecretKey) Raw() *RawSecretKey {
+	return raw
+}
+
+// Clone makes a clone of the secret key.
+func (raw *RawSecretKey) Clone() *RawSecretKey {
+	return &RawSecretKey{CipherData: raw.CipherData.Clone()}
+}
+
+// UnmarshalText converts the raw secret key back from a text marshaling.
+func (raw *RawSecretKey) UnmarshalText(text []byte) error {
+	raw.CipherData = &CipherData{}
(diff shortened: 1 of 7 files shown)


Gilthoniel:
First reduced PR for #587

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// PublicKey represents one of the two sides of an asymmetric key pair
+// which can be safely shared publicly.
+type PublicKey interface {
+	Nameable
+
+	fmt.Stringer
+
+	Raw() *RawPublicKey
```
tharvik: Is there a use case for returning nil?

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// SecretKey represents one of the two sides of an asymmetric key pair
+// which must remain private.
+type SecretKey interface {
+	Nameable
+
+	fmt.Stringer
+
+	Raw() *RawSecretKey
```
tharvik: ret nil?

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// Signature represents a signature produced using a secret key and
+// that can be verified with the associated public key.
+type Signature interface {
+	Nameable
+
+	fmt.Stringer
+
+	Raw() *RawSignature
```
tharvik: ret nil?

ciphersuite/ed25519.go
```diff
@@ -0,0 +1,221 @@
...
+
+const errNotEd25519CipherSuite = "invalid cipher suite"
+const errInvalidBufferSize = "invalid buffer size"
+
+// Ed25519CipherSuiteName is the name of the cipher suite that is using Ed25519 as
+// the signature algorithm.
+const Ed25519CipherSuiteName = "ED25519_CIPHER_SUITE"
+
+// Ed25519PublicKey is the public key implementation for the Ed25519 cipher suite.
+type Ed25519PublicKey struct {
```
tharvik: Why using a struct and not the type directly?

ciphersuite/ed25519.go
```diff
@@ -0,0 +1,221 @@
...
+	data ed25519.PublicKey
+}
+
+// Name returns the name of the cipher suite.
+func (pk *Ed25519PublicKey) Name() Name {
+	return Ed25519CipherSuiteName
+}
+
+func (pk *Ed25519PublicKey) String() string {
+	return hex.EncodeToString(pk.data)
```
tharvik: Why stringify (and not the same format as the RawPublicKey)?

tharvik: And some others in the same file.

ciphersuite/ed25519.go
```diff
@@ -0,0 +1,221 @@
...
+
+	if !ed25519.Verify(publicKey.data, msg, signature.data) {
+		return xerrors.New("signature not verified")
+	}
+
+	return nil
+}
+
+func (s *Ed25519CipherSuite) unpackPublicKey(pk PublicKey) (*Ed25519PublicKey, error) {
+	if data, ok := pk.(*RawPublicKey); ok {
```
tharvik: Why the casting? That looks like try to cope with library misuse; simply fail looks right

tharvik: and related below

ciphersuite/registry.go
```diff
@@ -0,0 +1,159 @@
...
+		// Cipher suite already registered so we return it so it can be reused.
+		return suite
+	}
+
+	cr.ciphers[name] = suite
+
+	return suite
+}
+
+func (cr *Registry) get(name Name) (CipherSuite, error) {
```
tharvik: value receiver

tharvik: and most of `Registry` (except `RegisterCipherSuite`)

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,304 @@
...
+	CipherName Name
+}
+
+// Name returns the name of the cipher suite compatible with the data
+// contained in the raw structure.
+func (d *CipherData) Name() Name {
+	return d.CipherName
+}
+
+func (d *CipherData) String() string {
```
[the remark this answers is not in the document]
Gilthoniel: There are a lot logs in Cothority mostly that are using a string representation of the point for incoming messages so this would be the replacement. True for the prefix though.

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,304 @@
...
+	return nil
+}
+
+// RawPublicKey is a raw data structure of a public key implementation.
+type RawPublicKey struct {
+	*CipherData
+}
+
+// NewRawPublicKey returns an instance of a public key.
+func NewRawPublicKey(name Name, data []byte) *RawPublicKey {
```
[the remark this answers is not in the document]
Gilthoniel: I'll see with the other engineers but some guidelines tell you to use the pointer receiver everywhere when some need for consistency.

ciphersuite/ciphersuite.go
Gilthoniel: Not, it's related to the pointer receiver thing.

ciphersuite/ed25519.go
Gilthoniel: You mean the ed25519 type ? Because I need to implement our abstraction that is used in Onet.

ciphersuite/ed25519.go
Gilthoniel: Good point

ciphersuite/ed25519.go
Gilthoniel: Because at this moment we don't use network messages and data structures in a separate way so I made the raw elements to be accepted as valid abstraction. The cipher suite must then convert them. It will very likely change when we will switch to the official protobuf.

ciphersuite/ciphersuite.go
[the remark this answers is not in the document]
tharvik: So that's mainly used to differentiate two keys, not ensuring uniqueness, so we can use a reduced version, such as the `$name:$few-first-hex-of-data`.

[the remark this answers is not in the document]
Gilthoniel: Yes, that's what I did. Actually reducing the hex value is also a good idea.

ciphersuite/ed25519.go
tharvik: Ho, okay, I'm discovering a bit more of golang, thanks

ciphersuite/ed25519.go
tharvik: TODO it or log it then?

ciphersuite/registry.go
```diff
@@ -0,0 +1,160 @@
...
+// UnpackPublicKey takes generic cipher data and tries to convert it
+// into a public key of the associated implementation. The cipher suite
+// must be registered beforehand.
+func (cr *Registry) UnpackPublicKey(raw *RawPublicKey) (PublicKey, error) {
+	c, err := cr.get(raw.Name())
+	if err != nil {
+		return nil, xerrors.Errorf("cipher suite: %v", err)
+	}
+
+	pk, err := c.PublicKey(raw)
```
nkcr: Can't we pass only `raw.Data`?  At this stage we are sure to have the correct cipher suite, so the check `raw.Name() != s.Name()` inside the `PublicKey()` function looks superfluous. Or I am missing a case?

ciphersuite/registry.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, xerrors.Errorf("signing: %v", err)
+	}
+
+	return sig, nil
+}
+
+// Verify takes a public key, a signature and a message and performs a verification
+// that will return an error if the signature does not match the message. It
+// will also return an error if the cipher suite is unknown.
+func (cr *Registry) Verify(pk PublicKey, sig Signature, msg []byte) error {
```
nkcr: just a thought, shouldn't we differentiate the case where the signature is incorrect from the one where an error occurred in the program? It feels a bit strange that the verification outputs an error while it only says that a signature is invalid.
Maybe it's better to rename the function `IsValid(...)` and return a `(bool, error)`. IDK

nkcr (changes requested):
I like this ciphersuite package, thanks for the nice job! :) 
Nice unit tests too 👍

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,306 @@
...
+	"encoding/hex"
+	"fmt"
+	"io"
+
+	"golang.org/x/xerrors"
+)
+
+// encodedNameLengthSize defines the size in bytes of the name length
+// when marshaling cipher data.
+const encodedNameLengthSize = 32 / 8
```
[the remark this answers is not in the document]
Gilthoniel: Something I wanted to discuss. 8 or 16 bits integer is probably big enough for a cipher name.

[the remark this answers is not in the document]
Gilthoniel: We'll see tomorrow morning

ciphersuite/ciphersuite.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// WriteTo implements the io.WriteTo interface so that the cipher
+// data can be written into any standard writer (e.g. hash).
+func (d *CipherData) WriteTo(w io.Writer) (n int64, err error) {
+	var size int
+	size, err = w.Write([]byte(d.Name()))
+	n += int64(size)
+	if err != nil {
+		return n, xerrors.Errorf("writing name: %v", err)
```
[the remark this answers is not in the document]
Gilthoniel: Well the fact that we are wrapping the errors create a chain that should explain what happened. Like here a calling function would have something like `marshaling cipher data: writing name: ...` so you know what went wrong.

But I'm curious to know what you would have written.

ciphersuite/ed25519.go
```diff
@@ -0,0 +1,220 @@
+package ciphersuite
+
+import (
+	"io"
+
+	"golang.org/x/crypto/ed25519"
+	"golang.org/x/xerrors"
+)
+
+const errNotEd25519CipherSuite = "invalid cipher suite"
```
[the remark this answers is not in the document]
Gilthoniel: No it was only to avoid repetitive string values but for the moment I don't think it's necessary. We can always change that later.

ciphersuite/ed25519.go
```diff
@@ -0,0 +1,220 @@
...
+
+	if publicKey, ok := pk.(*Ed25519PublicKey); ok {
+		return publicKey, nil
+	}
+
+	return nil, xerrors.New("wrong type of public key")
+}
+
+func (s *Ed25519CipherSuite) unpackSecretKey(sk SecretKey) (*Ed25519SecretKey, error) {
+	if data, ok := sk.(*RawSecretKey); ok {
```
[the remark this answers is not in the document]
Gilthoniel: I think this kind of ok conversion is quite frequent in Go

ciphersuite/ed25519.go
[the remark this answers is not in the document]
tharvik: btw, if you want to have it public, have it as an `error`, so that `errors.Is` can be used directly.

ciphersuite/ciphersuite.go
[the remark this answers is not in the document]
tharvik: Or you can use `binary.Uvarint`, to pack tighter, reduce code complexitiy and have unlimited size.

ciphersuite/ciphersuite.go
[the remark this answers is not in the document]
nkcr: I look where the error comes from and explain it:

`"failed to write Name as bytes: %v", err`

Gilthoniel:
On hold during the TLS related investigation.

ciphersuite/ciphersuite.go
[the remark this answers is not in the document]
ineiti: That was my guideline - but I have been proven wrong ;) We were trying to get away from this for `v3`, but then couldn't. So if you still think it's wrong to always use pointer structs in `New` methods, please go ahead and open an issue for `v4` to actually return the more appropriate structures.

One place where it will collide with a lot of code is the 'old' protobuf-structures, where the pointer indicates an optional value.

ciphersuite/ed25519.go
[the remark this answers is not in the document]
ineiti: I had to get used to them, too... But I saw even worse (don't remember where):

```
if ok := some_method; !ok {
}
```

ciphersuite/registry.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// NewRegistry creates a new empty registry.
+func NewRegistry() *Registry {
+	return &Registry{
+		ciphers: make(map[Name]CipherSuite),
+	}
+}
+
+// RegisterCipherSuite stores the cipher if it does not exist. It returns the
```
cgrigis:
```suggestion
// RegisterCipherSuite stores the cipher if it does not exist. It returns
```

ciphersuite/registry.go
```diff
@@ -0,0 +1,160 @@
...
+// the suite stored for this name if it already exists, or it returns the
+// provided suite.
+func (cr *Registry) RegisterCipherSuite(suite CipherSuite) CipherSuite {
+	name := suite.Name()
+	if suite := cr.ciphers[name]; suite != nil {
+		// Cipher suite already registered so we return it so it can be reused.
```
cgrigis: Isn't it surprising for the caller to receive a suite different that the provided one, with no error?

nkcr:
This PR is celebrating it's first year tomorrow 🎂  😰
Should we consider merging it ? It would require a bit of effort to address some of the comments, anyone's up to do that ?

ineiti:
I propose to close this issue without merging for the following reasons:
- I think it solves the wrong problem - see my comment in the related issue.
- there already is a way to have independent keypairs for each service. So if anything, this interfaces should be extended, and not a new added.
- it was originally thought to help implement Drynx / Spindle from the LDS, but finally they used their own keypair implementation

PR #596, opened 2019-10-31
Waiting on Local.CloseAll

ineiti:
A lot of tests using ByzCoin have a bad timing when they call local.CloseAll,
because they have a waiting call to 'AddTransactionAndWait', which then needs to
timeout. Adding a non-failing 'WaitDone' at the beginning of Local.CloseAll
removes most of those problems.


jeffallen:
I currently not planning to take this for v3, since it seems to be about stability during dev work.

PR #603, opened 2019-11-26
Use the routine dispatcher to deserve requests asynchronously

Gilthoniel:
In situation where a node is processing a request that needs interactions with other nodes, it
can be stuck as the blocking dispatcher only allows one at a time. This switches to the routine
dispatcher that allows multiple requests to be processed.

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -67,7 +67,7 @@ func NewRouter(own *ServerIdentity, h Host) *Router {
 		ServerIdentity:          own,
 		connections:             make(map[ServerIdentityID][]Conn),
 		host:                    h,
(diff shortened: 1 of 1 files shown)


Gilthoniel:
@cba33 has issues during simulation as her init phase requires interaction with the nodes and thus the root was stuck waiting for replies to be processed.

ineiti:
IIRC, this was half on purpose, to avoid some race-conditions when the dispatcher is called twice for the same message. But probably it never really prevented races... If there are more flaky tests in cothority, this might be it...

jeffallen:
How does this change perform when integrated into Cothority?

Gilthoniel:
Running the tests locally was alright.

PR #613, opened 2020-01-16
Dbadmin

ineiti:
Problem statement: While wanting to separate my conode to restrict it to do only byzcoin stuff, but keeping the data of the personhood-parties, I wanted to take the data from the personhood-service and put it in the new conode.

My solution:
- `dbadmin inspect` - to see which buckets are available
- `dbadmin extract` - to copy certain buckets to a new db

So in my case, I could do:
```bash
dbadmin extract --source old_conode.db --destination new_conode.db Personhood.*
```


nkcr:
Since this tool is not directly related to Onet, shouldn't we release it in a new repo?

ineiti:
I don't like proliferation of repos. It is directly related to onet, as such I think it is a good place. But if you prefer to have it elsewhere, I'll add it to https://github.com/c4dt/byzcoin.

nkcr:
I don't know what is best and feel the same about creating new repos.
The rational for my comment is that we are sometime using bolt for other projects. Therefore, having a general purpose cli for bolt could be useful and should not be tight to Onet.

go.sum
```diff
@@ -1,5 +1,7 @@
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
+github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d h1:U+s90UTSYgptZMwQh2aRr3LuazLJIa+Pg3Kc1ylSYVY=
```
jeffallen: I'm wondering where these came from, perhaps a "go get" in this directory, when it should have been in $HOME?

app/libtest.sh
```diff
@@ -16,9 +16,23 @@ APPDIR=${APPDIR:-$(pwd)}
 APP=${APP:-$(basename $APPDIR)}
 # Name of conode-log
 COLOG=conode
+# Base port when generating the configurations
```
jeffallen: Are all of these changes just from a sync with cothority/libtest.sh?

ineiti: Yes, that's a verbatim copy of `cothority/libtest.sh`. Not sure how you want to handle this, but I needed `libtest.sh`, so instead of copying it in `onet/dbadmin`, I updated the `onet/app/libtest.sh`.

go.sum
ineiti: That came with the urfave-cli.

PR #616, opened 2020-01-22
Updates the ServerHTTP function to handle streaming from the client

nkcr:
\+ some refactoring


ineiti:
I'm not sure it's not asking for problems the way you update `tun` in the go-routine... It seems to be racy to me. 

Also, does it make sense to send commands once you opened a stream? The way you handle it now, if a command is sent, a new stream is created. But is the old one closed? It doesn't seem. So every command creates a new stream, correct?

If I'm correct that you're opening multiple streams at the same time, I have no idea how websocket handles it. My proposition is to either

- close the stream before sending another command
- close the stream when a command is sent

But as I don't know what you're doing on the client side, I don't know if either of these make sense.

nkcr:
> I'm not sure it's not asking for problems the way you update `tun` in the go-routine... It seems to be racy to me.

Ok, I removed the update of the variable

> Also, does it make sense to send commands once you opened a stream?

That allows the client to send additional message without needing to close the stream and re-open one, which is a feature of WebSocket.

> The way you handle it now, if a command is sent, a new stream is created. But is the old one closed? It doesn't seem. So every command creates a new stream, correct?

No. In the way I implemented the service, the client gets back a streamID when it sends the first message. If it's the first message then the service creates a stream. Then, after the first message sent to the service, the client can re-send a message with the streamID, and the service will re-use the already created stream.

service.go
```diff
@@ -47,6 +47,8 @@ type Service interface {
 	// should be used. The returned error will be formatted as a websocket
 	// error code 4000, using the string form of the error as the message.
 	ProcessClientRequest(req *http.Request, handler string, msg []byte) (reply []byte, tunnel *StreamingTunnel, err error)
+	ProcessClientRequest2(req *http.Request, path string, buf []byte, furtherInputs chan []byte) ([]byte, *StreamingTunnel, error)
+	IsStreaming(path string) (bool, error)
```
ineiti: Adding these two methods to the `Service` interface breaks all the existing services!

Another way to do it is to define a `BidirectionalStreamer interface` with only the `ProcessClientRequest2` in it (with a better name). Then, instead of doing `if s.IsStreaming` you can do 

```
if bisi, ok := s.(*BidirectionalStreamer){
  bisi.ProcessClientRequest2....
} else {
  s.ProcessClientRequest
}
```

processor.go
```diff
@@ -442,6 +442,125 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
...
+				panic("no such channel index")
+			}
+			// We don't add a way to explicitly stop the
+			// go-routine, otherwise the service will
+			// block. The service should close the channel
+			// when it has nothing else to say because it
+			// is the producer. Then this go-routine will
+			// be stopped as well.
+		}
+		log.LLvl1(">>>>>> return from the first go routine")
```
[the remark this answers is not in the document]
nkcr: good to know, thanks 👍

processor_test.go
```diff
@@ -195,6 +195,70 @@ func TestServiceProcessor_ProcessClientRequest_Streaming(t *testing.T) {
...
+	for i := 0; i < n; i++ {
+		buf := <-tun.out
+		val := &testMsg{}
+		require.Nil(t, protobuf.Decode(buf, val))
+		require.Equal(t, val.I, int64(n))
+	}
+
+	close(tun.close)
+
+	// Would normall be done inside the service, but we would then need to be
```
jeffallen: s/normall/normally/

Also: I use unit tests as documentation on how to use a feature. So it would be helpful to show readers here exactly what is expected of them, should they choose to call RegisterStreamingHandler. This unit test should do it precisely the way users of Onet should do it, without taking advantage of the unit test context.

websocket.go
```diff
@@ -278,60 +278,101 @@ outerReadLoop:
...
-					break
+
+		isStreaming := false
+		bidirectionalStreamer, ok := s.(BidirectionalStreamer)
+		if ok {
+			isStreaming, err = bidirectionalStreamer.IsStreaming(path)
+			if err != nil {
+				log.Errorf("failed to check if it is a streaming "+
+					"request %s/%s: %+v", t.serviceName, path, err)
+				continue
```
jeffallen: Each of these continue/breaks could use a comment explaining the reasoning for why it has to be that. I suspect that this part of the code is pretty lightly (or not at all?) covered by unit tests. What does code coverage say?

This is an important place to be careful, because getting the outer loop into an endless loop would be very bad here: hung grouting in a spin loop, possibly. (Been there, done that, got the coredump.)

websocket_test.go
```diff
@@ -564,11 +564,11 @@ func TestWebSocketTLS_Error(t *testing.T) {
 	require.NotEqual(t, "", log.GetStdErr())
 }
 
-// TestWebSocket_Streaming performs 3 test cases.
+// TestWebSocket_Streaming_normal performs 3 test cases.
```
jeffallen: I wonder if this would be a good time to separate these out into 3 test cases? I'm also not clear why the rest for your new code is in processor_test.go and not here. (Not saying you should change, just want to understand your reasoning.)

processor.go
```diff
@@ -442,6 +442,123 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
 	return
 }
 
+// ProcessClientStreamRequest allows clients to push multiple messages
+// asynchronously to the same service with the same connection. Unlike in
+// ProcessClientRequest, we take a channel of inputs that can be filled and will
+// subsequently call the service with any new messages received in the channel.
+func (p *ServiceProcessor) ProcessClientStreamRequest(req *http.Request, path string,
```
Gilthoniel: Would it be possible to use this function inside `ProcessClientRequest` so that we reduce the code duplication ?

processor.go
```diff
@@ -442,6 +442,123 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
...
+		}
+	}()
+
+	// This goroutine listens on any new messages from the client and executes
+	// the request. Executing the request should fill the service's channel, as
+	// the service will use the same chanel for further requests.
+	go func() {
+		for {
+			select {
+			case buf := <-clientInputs:
```
Gilthoniel: If the channel is closed by the caller, it might panic in the decode function.

processor.go
```diff
@@ -442,6 +442,123 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
...
+// ProcessClientRequest, we take a channel of inputs that can be filled and will
+// subsequently call the service with any new messages received in the channel.
+func (p *ServiceProcessor) ProcessClientStreamRequest(req *http.Request, path string,
+	clientInputs chan []byte) ([]byte, *StreamingTunnel, error) {
+
+	outChan := make(chan []byte, 100)
+	buf := <-clientInputs
+	mh, ok := p.handlers[path]
+
+	reply, stopServiceChan, err := func() (interface{}, chan bool, error) {
```
Gilthoniel: Is there an actual need/benefit of using a function here ?

processor.go
```diff
@@ -442,6 +442,123 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
...
+		return callInterfaceFunc(mh.handler, msg, mh.streaming)
+	}()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// This goroutine is responsible for listening on the service channel,
+	// decoding the messages and then forwarding them to the streaming tunnel,
+	// which should then forward the message to the client.
+	go func() {
```
Gilthoniel: Instead of running the go routine right away thus forcing you to listen for the client input (which could hang, be closed, etc.. so more stuff to manage in two places), why don't you start by listening client inputs and spinning up a go routine for the service processing ? You can wait for completion if you want to be sure to be synchronous.

websocket_test.go
nkcr: Most of the new work is the implementation of `ProcessClientStreamRequest()` in processor.go, where the changes in websocket.go are mainly refactoring. So I primarily wanted to unit test this new function. 
But yes, let's also update websocket_test.go. Increasing test coverage won't hurt.

websocket.go
nkcr: Agreed. It follows the same error handling as what has already been implemented: if err is not nil then the loop ends. 
(Coverage: ~85%)

processor.go
nkcr: surprisingly not: if the buffer is empty, the `DecodeWithConstructors` doesn't return an error, it just has no effect.

nkcr: the annoying thing is that it loops until the stopServiceChan is closed.

nkcr: so, should we say that as soon as there is an empty buffer we exit ?

Gilthoniel: Or you can slightly change the API to not return the streaming tunnel but only the out channel and the caller has to close the client inputs channel when he's done.

nkcr: ok, I did that and it make total sense. We should be careful though, it's starting to look like real streaming...;) 

I don't know if we can keep it like that (probably not), but I then updated the `ProcessClientRequest` to reject streaming request, saying that it should instead use `ProcessClientStreamRequest`. Note that this does not affect the behavior of websocket.go

processor.go
nkcr: I removed the handling of streaming transaction from ProcessClientRequest, which removes most of the duplication.

processor.go
nkcr: except if we need the code to be harder to read I guess no ;)

processor.go
nkcr: If I understand you well, this is already the case: In L453 I wait for the client to send a first message, then I launch the go routines.

nkcr: Updated so that each new request from the client creates a new goroutine.

nkcr:
@Gilthoniel thanks for the review, all points have been adressed, and tests for byzcoin/java pass.

processor.go
```diff
@@ -442,11 +443,133 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
...
+					log.Error(xerrors.Errorf("decoding: %v", err))
+					return
+				}
+
+				reply, stopServiceChan, err = callInterfaceFunc(mh.handler, msg, mh.streaming)
+				if err != nil {
+					log.Error(err)
+					if stopServiceChan != nil {
+						close(stopServiceChan)
+					}
```
Gilthoniel: missing a return there ?

processor.go
```diff
@@ -442,11 +443,133 @@ func callInterfaceFunc(handler, input interface{}, streaming bool) (intf interfa
...
+							return
+						}
+						if chosen == 0 {
+							// Send information down to the client.
+							buf, err = protobuf.Encode(v.Interface())
+							if err != nil {
+								log.Error(err)
+								return
+							}
+							outChan <- buf
```
Gilthoniel: I think you could end up in a situation where two go routines are running, one closes the out channel and the second tries to write into it which will end up in a panic.

processor.go
Gilthoniel: Shouldn't the return be for any error ? Now it's returning only if the service returns a channel.

nkcr: ugh... absolutely!

PR #623, opened 2020-03-13
Adding tracing services to Onet

### LINKED ISSUE #618: Enable Observability
The current network sometimes crashes, and we're not really sure why. A simple system has been put into place using graylog, but something more elaborate debugging needs to be done.

As a first step, the onet/log could be used to send traces to a central system like

https://opentracing.io/
https://zipkin.io/
https://prometheus.io/
https://honeycomb.io/

ineiti:
Instead of simply logging message to the console, syslog, or graylog,
it is much better to do actual tracing of your code.
To avoid having to add tracing messages throughout the code, this PR
proposes a new onet-logger that converts the current log-calls to
tracing-magick.

Currently there are two tracers written:
- honeycomb.io
- simulation
But the framework is open enough that other tracing systems should be
pluggable in the same way.

This code is currently tested on c4dt's 2nd conode, and it seems to
work nicely.

Closes #618


ineiti:
TODO (please add others):
- check why the structures only show up as strings on honeycomb.io

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+//  let you trace where you could/should add a TRACING_ENTRY_POINT
+//   - TRACING_CREATE_SINGLE_SPANS -
+//   - TRACING_ENTRY_POINTS - a "::" separated list of entry points that can
+//  be used to refine the tracing.
+//  The name of the entry points are the same as given by
+//  TRACING_PRINT_SINGLE_SPAN
+//   - TRACING_DONE_MSGS - a "::" separated list of done msgs that will close
+//  the started trace. This is due to the simulation of actual traces,
+//  because we can't really know when the trace should end.
+func (l *TraceLogger) AddEnvironment() {
```
tharvik: IMO, it should fail when parsing the env, rather than warning.

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+		fmt.Printf("Creating single trace for '%s' from:\n",
+			msg)
+		for i, se := range ses {
+			if i >= l.PrintSingleSpans {
+				break
+			}
+			fmt.Println("\t", se.pkgMethod())
+		}
+	}
+	if l.NoSingleSpans == false {
```
tharvik:
```suggestion
	if !l.NoSingleSpans {
```

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+	}
+	if l.PrintSingleSpans > 0 {
+		fmt.Printf("Creating single trace for '%s' from:\n",
+			msg)
+		for i, se := range ses {
+			if i >= l.PrintSingleSpans {
+				break
+			}
+			fmt.Println("\t", se.pkgMethod())
```
tharvik: Debug?

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+		if se.checkEntryPoint(l.entryPoints) != "" {
+			var child *spanWrapper
+			l.traces[se.methodStr()], child =
+				l.newTrace(context.TODO(), "", ses[i:]...)
+			child.log(lvl, msg)
+			return l.traces[se.methodStr()], child
+		}
+	}
+	if l.PrintSingleSpans > 0 {
+		fmt.Printf("Creating single trace for '%s' from:\n",
```
tharvik: Debug?

tracing/logger_test.go
```diff
@@ -0,0 +1,73 @@
...
+	log.AddUserUninterestingGoroutine("writeLoop")
+	log.AddUserUninterestingGoroutine("worker")
+	local := onet.NewLocalTest(tSuite)
+	server := local.GenServers(1)[0]
+	if server.UnauthOk {
+		log.Print("OK to unauth")
+	}
+	time.Sleep(time.Second)
+	defer local.CloseAll()
+	defer beeline.Close()
```
tharvik: `defer` at the end?

tracing/logger_test.go
```diff
@@ -0,0 +1,73 @@
...
+	os.Setenv("HONEYCOMB_API_KEY", "API_KEY:Dataset")
+	log.AddUserUninterestingGoroutine("runtime_pollWait")
+	log.AddUserUninterestingGoroutine("writeLoop")
+	log.AddUserUninterestingGoroutine("worker")
+	local := onet.NewLocalTest(tSuite)
+	server := local.GenServers(1)[0]
+	if server.UnauthOk {
+		log.Print("OK to unauth")
+	}
+	time.Sleep(time.Second)
```
tharvik: Whut?

tracing/service/service.go
```diff
@@ -0,0 +1,53 @@
...
+func newTracer(c *onet.Context) (onet.Service, error) {
+	if loggerCounter > 0 {
+		log.Warn("can only start one service for tracing")
+		return nil, nil
+	}
+	loggerCounter++
+	tl, err := tracing.NewHoneycombLoggerFromEnv()
+	if err != nil {
+		log.Errorf("couldn't start honeycomb: %v", err)
+		return nil, nil
```
tharvik:
```suggestion
		return nil, err
```

tracing/service/service.go
```diff
@@ -0,0 +1,53 @@
...
+
+type tracer struct {
+	*onet.ServiceProcessor
+	tl *tracing.TraceLogger
+}
+
+func newTracer(c *onet.Context) (onet.Service, error) {
+	if loggerCounter > 0 {
+		log.Warn("can only start one service for tracing")
+		return nil, nil
```
tharvik:
```suggestion
		return nil, errors.New("can only start one service for tracing")
```

tracing/trace.go
```diff
@@ -0,0 +1,272 @@
...
+type Span interface {
+	AddField(key string, val interface{})
+	CreateChild(ctx context.Context) (context.Context, Span)
+	Send()
+}
+
+// traceWrapper keeps track of the stackEntries and creates / sends Spans as
+// needed.
+type traceWrapper struct {
+	ctx     context.Context
```
tharvik: `context.Context` shouldn't be stored in struct.

tracing/trace.go
```diff
@@ -0,0 +1,272 @@
...
+	if sp.child != nil {
+		sp.child.done()
+	}
+	_, sLog := sp.hcSpan.CreateChild(sp.ctx)
+	msgStruct := struct {
+		Lvl int
+		Msg string
+	}{lvl, msg}
+	// This value is unfortunately not exported from onet/log/lvl.go
+	if lvl > -17 {
```
tharvik: Magic?

tracing/honeycomb.go
```diff
@@ -0,0 +1,67 @@
...
+// The HoneyComb implementation is mostly a wrapper to convince go that it's
+//fine returning structures as interfaces.
+// A Honeycomb logger can either be initialised using
+//  - NewHoneycombLogger if you know the api key and the dataset
+//  - NewHoneycombLoggerFromEnv if you want the user to chose them using an
+//    environment variable.
+
+type honeycombTrace struct {
+	*trace.Trace
+	ctx context.Context
```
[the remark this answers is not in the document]
ineiti: well here I must ;) It's all about the simulating of an actual trace...

ineiti:
> As I understood it, tracing is quite different from logging. Logging show status/progress at each step, tracing try to measure a span of execution.

Yes, and that's the whole point of this PR: convert the current logging into a tracing, using the fact that golang's stack-traces are detailed enough to separate the different traces.

So whenever a `log` comes in, the `Trace` structure looks it up and checks whether that trace already exists. Then it inserts it correctly in the tree, simulating calls from methods to another. This is why it needs to keep track of `context.Context`.

The alternative is to actively rewrite onet and cothority to use a tracing framework...

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+	doneMsgs         []string
+	traces           map[string]*traceWrapper
+	logMutex         sync.Mutex
+}
+
+type newTrace = func(context.Context, string) (context.Context, Trace)
+
+// Log implements the Logger interface.
+func (l *TraceLogger) Log(level int, msg string) {
+	msg = strings.TrimSpace(msg)
```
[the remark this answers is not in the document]
ineiti: Old line - will be removed. Before these changes, the loggers received a line ending in `\n`. Now I changed it, using that awful hack of using `fmt.Sprintln` to have space-separated arguments.

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+
+// AddEntryPoints takes all given entry points and adds them to the internal
+// list.
+// Empty entry points are discarded.
+func (l *TraceLogger) AddEntryPoints(eps ...string) {
+	for _, ep := range eps {
+		if len(ep) > 0 {
+			l.entryPoints = append(l.entryPoints, ep)
+		}
+	}
```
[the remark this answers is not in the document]
ineiti: Because the `entryPoints` and `doneMsgs` are somewhat crucial I prefer having them private and then checking for empty strings before adding them.
Also, not failing here allows for a simpler usage for environment-variables by simply passing the `strings.Split` result.

tracing/logger.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// AddDoneMsgs takes all given done messages and adds them to the internal list.
+// Empty done messages are discarded.
+func (l *TraceLogger) AddDoneMsgs(dms ...string) {
+	for _, dm := range dms {
+		if len(dm) > 0 {
+			l.doneMsgs = append(l.doneMsgs, dm)
+		}
+	}
```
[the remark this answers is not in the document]
ineiti: Same reasoning.

tracing/logger.go
ineiti: I put this on purpose on `== false`, because the `No` is an inversion... But if you think it doesn't help read the code, I'll use the `!`, which, btw, is very similar to the `l`, and so might be overlooked...

tracing/logger.go
ineiti: no, as this is in a `Logger`, I cannot use `log`...  So this is how the tracing system prints potential `EntryPoints` to the user. This is explained in the `README.md`

tracing/logger.go
ineiti: no, no `log` in `Logger`...

tracing/logger_test.go
ineiti: oups - that was supposed to go on afterwards... But then I wrote the `simulCollector` and didn't needed it.

tracing/logger_test.go
ineiti: Yes, bad. It's a test that allows to check whether the connection to honeycomb actually works or not. It's very basic, but did its job. I can also remove the whole test, but thought it's helpful if needed later.

As honeycomb sends the messages asynchronously, there is no guarantee that `.Send()` is done when it returns. So you need to wait...

It's a `skip`ped test with a reason ;)

tracing/service/service.go
ineiti: Changed the error-handling of `NewHoneycombLoggerFromEnv`...

tracing/service/service.go
ineiti: hmm - hard call. It's not that the system is unusable. It's just that the user asked to trace in a simulation with more than one node running in the same process - which might make sense. But which is not standard.

tracing/trace.go
ineiti: Yes, explained in the comment above...

tracing/logger.go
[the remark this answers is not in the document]
tharvik: In `AddEnvironment`, you check for the content of `TRACING_PRINT_SINGLE_SPAN` and `TRACING_CREATE_SINGLE_SPANS`, shouldn't be the check for empty values be put there instead?
Maybe, someone wants to trace every function, thus using `""` (or `"."` btw) as a valid entrypoint.

tracing/logger.go
tharvik: We can have both, by using a `logger` instead of `l`.

gnarula:
> So whenever a `log` comes in, the `Trace` structure looks it up and checks whether that trace already exists. Then it inserts it correctly in the tree, simulating calls from methods to another. This is why it needs to keep track of `context.Context`.

I might be wrong here but IMHO logs are something which are isolated to a single node and traces  "span" across multiple nodes. Case in point, the hierarchy in the example at https://www.jaegertracing.io/img/trace-detail-ss.png.

IIRC, tracing libraries usually keep the span info in a `context.Context` and propagate it for all network requests. I'm not sure if we're doing the same here because `getTraceSpan` is only looking at the method signature. Am I missing something?

In other words, how would a message being handled by tree node at level N know the span info about its parent at level N - 1?

ineiti:
> In other words, how would a message being handled by tree node at level N know the span info about its parent at level N - 1?

That's the limit of the solution proposed here. Theoretically one could push it further and use the protocol-ID to merge traces. But then I think it would be difficult to find a frontend that can visualize a BFT-BLSCosi signature. Or, even worse, a full trace starting from a `ClientTranaction` request to the final block...

So the scope of this PR is to have a better log-aggregation that goes somewhat towards tracing. The `traceWrapper` structure keeps track of the 'simulated' `context.Context` structure. What I'm currently struggling with are the go-routines, as they are difficult to track through the string returned by `runtime.Stack`.

ineiti:
OK, without adding extra information in the code this will be very difficult. I tried to get an automatic detection of which go-routines go together, but this is not something that go supports. So you have to scrape it together using `runtime.Stack` and stuff like that, which makes it awful.

I think I'll add a `log.TraceID` anyway where you can set an ID, if you have one. Currently the only ID that is available is `RoundID` from the `onet.Token`, which is used in protocols. But it is not ideal - the best would be of course to replace all the `log.` messages with _real_ tracing messages. But that looks like a _lot_ of work :(

ineiti:
`log.TraceID` is added and seems to work nicely. Now I can for example follow a `blscosi` signature through the different phases. It's not too pretty, but not too bad, neither.

It's a bit finnicky to set up with putting the `log.TraceID` in the right places.

ineiti:
OK, this looks somewhat usable. I'm waiting for @tharvik and @cgrigis to play around with honeycomb.io, and if they have a not-too-difficult proposition to be included, I'll do it.

Else this looks nearly final to me (except @tharvik's code suggestions, of course).

tracing/logger.go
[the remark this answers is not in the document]
ineiti: The `SINGLE_SPANS` are more of a mapping of log-messages to the tracing framework, and not really useful. As such they are not an `entry point`, as they will not be able to receive other spans. So it's two different things.

tracing/logger.go
```diff
@@ -0,0 +1,297 @@
...
+				return
+			}
+		}
+	}()
+}
+
+// AddEnvironment reads the environment variables defined to initialize the
+// variables.
+// The following environmental variables are available:
+//   - TRACING_PRINT_SINGLE_SPAN - output a stack trace of single spans to
```
cgrigis: In the README it is referred to as `TRACING_PRINT_SINGLE_SPANS` (with an S).

tracing/logger.go
```diff
@@ -0,0 +1,297 @@
...
+//  let you trace where you could/should add a TRACING_ENTRY_POINT
+//   - TRACING_CREATE_SINGLE_SPANS - whenever there is no entry point found,
```
cgrigis: This one is not documented in the README.

tracing/trace.go
cgrigis: It would be helpful to document it refers to `lvlPanic` though. :)

tracing/trace.go
```diff
@@ -0,0 +1,385 @@
...
+	file      string
+	pkgPath   string
+	method    string
+	params    string
+	traceID   string
+	createdBy bool
+}
+
+var stackFileLine = regexp.MustCompile(
+	"\\s*(.*):([0-9]*)( \\+0x[a-f0-9]*$)?\\s*")
```
cgrigis: Might be clearer to use a raw string, i.e. `` `\s*(.*):([0-9]*)( \+0x[a-f0-9]*$)?\s*` ``

tracing/trace.go
```diff
@@ -0,0 +1,385 @@
...
+var stackPkgMethod = regexp.MustCompile(
+	"\\s*(created by )?(.*)\\.(.*?\\w)(\\(.*\\))?\\s*$")
```
cgrigis: I might be wrong, but this kind of parsing feels brittle... are there any guarantees Go's stacktraces will keep the same format?

tracing/trace.go
ineiti: Thanks - I only discovered `` ` ` `` halfway through and then forgot to change the others ;)

tracing/trace.go
ineiti: Yes, unfortunately it was the only place I could find reliable data about which go-routine is running and the stack-trace. All other calls in `debug`, `runtime`, and `runtime/trace` didn't give me the information I needed. And most of those calls also just return a string and not an actual usable structure.

One thing I did not check is to create an actual trace with `trace.Start` and then evaluate the binary stream and parse it. But I'm not even sure this would be real-time or whether it is just written in junks.

When re-writing, the `trace` library has handy methods - but you'd need to add them throughout the code. And I was looking for a minimal invasive solution.

ineiti:
As soon as it passes the tests I think it's R4M. Except comments, of course.

ineiti:
Heisenbug - need to fix it first.

ineiti:
Found it - goroutines are of course not deterministic... Should be ready 4 merge now.

PR #626, opened 2020-03-23
websocket: getWSHostPort uses URL if available

tharvik:
Weirdly enough, when specifying an `URL` in a conode's config, it doesn't bind the WebSocket tunnel to it, but to the `Address` with its port incremented. This PR fixes that by actually taking `URL` into consideration if `!= ""`.

It will probably **break the servers using `URL` with unspecified port**, such as https://wookiee.ch/conode & https://conode.dedis.ch

### DIFF
--- a/websocket.go
+++ b/websocket.go
@@ -3,6 +3,7 @@ package onet
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -111,8 +112,7 @@ type WebSocket struct {
 	sync.Mutex
 }
 
-// NewWebSocket opens a webservice-listener one port above the given
-// ServerIdentity.
+// NewWebSocket opens a webservice-listener at the given si.URL.
 func NewWebSocket(si *network.ServerIdentity) *WebSocket {
 	w := &WebSocket{
 		services:  make(map[string]Service),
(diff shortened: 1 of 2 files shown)


ineiti:
Can you please check with @jeffallen and @Gilthoniel that this does not break the setup of the DEDIS node? I know they also have port 443, but I don't know how they solved it internally...

If it breaks their setup and they want to have it (somewhat) backward-compatible, pick one:
1. Add an additional `AddressWS` for the binding address that is independent of `URL` - full backward compatibility
2. Interpret the `URL` differently in `public.toml` (for calls of clients) and `private.toml` (where to bind) - somewhat backward compatibility

ineiti:
I think Jeff is AFK for some time - you can try to contact Gaylor on slack.

Gilthoniel:
I know the Javascript client is using URL correctly but looking at the changes, it looks like it wouldn't work for with the current setup without @tharvik 's changes. I wonder if we ever contacted the DEDIS conode with a Go client..

ineiti:
> I wonder if we ever contacted the DEDIS conode with a Go client..

Definitely - when trying to look at why the chain broke we did contact _all_ nodes, IIRC, it was already on port 443.

Gilthoniel:
Note that you can see on https://status.dedis.ch that port 443 is used when not specified. IIRC, we changed Onet to listen of the protocol port if not specified.

ineiti:
> Note that you can see on https://status.dedis.ch that port 443 is used when not specified. IIRC, we changed Onet to listen of the protocol port if not specified.

I'm not sure the DEDIS node actually listens on port 443. According to the tests @tharvik did, it does listen to port+1, which would be `7001`, and then the apache has a websocket-proxy that redirects incoming requests from 443 to 7001. Then again, I might have missed something. But I want to make sure that if we break your node, you know it beforehand, and know what to do ;)

tharvik:
> Could you add unit tests to make sure a server identity is correctly translated into a valid address ? With the different use cases and especially the ones you stated ?

Good idea, I added some.

PR #632, opened 2020-03-27
Fixes issue with simulation

### LINKED ISSUE #631: Simulation on mininet still not working
There is still an error when trying to use mininet: `start.py` is read-only, so the script cannot overwrite it. And it lacks the `x` attribute...

ineiti:
Closes #631

Moved all initialization of mininet to mininet.go.
Now correctly searches the path for all scripts that should be run.

### DIFF
--- a/simul/platform/mininet.go
+++ b/simul/platform/mininet.go
@@ -84,20 +84,58 @@ type MiniNet struct {
 	Tags string
 }
 
+func newMiniNet() (m *MiniNet, err error) {
+	cmd := exec.Command("go", "list", "-f", "{{.Dir}}",
+		"-m", "go.dedis.ch/onet/v3")
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, xerrors.Errorf("while getting platform-path: %v", err)
+	}
+	m = &MiniNet{
+		mininetDir: path.Join(strings.TrimSpace(string(out)), "simul",
+			"platform", "mininet"),
+	}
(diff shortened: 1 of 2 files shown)


simul/platform/mininet.go
```diff
@@ -248,11 +286,17 @@ func (m *MiniNet) Deploy(rc *RunConfig) error {
 	}
 
 	// Copy our script
-	err = app.Copy(m.deployDir, m.mininetDir+"/start.py")
+	err = app.Copy(m.deployDir, path.Join(m.mininetDir, "start.py"))
 	if err != nil {
 		log.Error(err)
 		return xerrors.Errorf("copying: %v", err)
 	}
+	// As the `start.py` is kept in the go-pkg directory now,
```
jeffallen: The go pkg dir should really not be touched at all. If they include the permissions in the source hashes, then this change would cause "go mod verify" to start failing, etc. Safer would be copy to a https://golang.org/pkg/io/ioutil/#TempFile, chmod, and defer a delete to clean it up.

But: if this works, and you're happy with it, we can go with it, and change it later if needed.

ineiti: This doesn't touch the go mod cache dir at all. The file from the go mod cache dir is copied to the `m.deployDir`, which is `$(pwd)/deploy`, and only `$(pwd)/deploy/start.py` is changed.

Of course I would _never_ dare touching the go mod cache dir (cough, cough)...

PR #634, opened 2020-04-14
Reducing the size of the honeycomb traces

ineiti:
The first version sent information about GC and memory along every trace.
As this information is quite heavy, it has been moved together with other
information to AddStats, which is only called once per minute.

Reviewers: one OR another

### DIFF
--- a/tracing/logger.go
+++ b/tracing/logger.go
@@ -11,11 +11,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/shirou/gopsutil/cpu"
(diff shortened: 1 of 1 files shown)


ineiti:
I'm still confused if I should merge the PR now ;) 'Officially' I only have merge-rights on cothority. Should I still merge? Am I overcautious?

nkcr:
good question, in my mind I was in the cothority mode but you're right, we haven't officially discuss the onet case ; )

PR #636, opened 2020-04-20
Fix logging with DEBUG_LVL=0

ineiti:
After the tracing addition, DEBUG_LVL=0 produced two outputs per
log-call.

This PR fixes it and adds a test to avoid regression.

### DIFF
--- a/log/loggers.go
+++ b/log/loggers.go
@@ -130,6 +130,10 @@ type stdLogger struct {
 }
 
 func (sl *stdLogger) Log(lvl int, msg string) {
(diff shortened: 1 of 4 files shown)


log/loggers.go
```diff
@@ -129,6 +130,11 @@ type stdLogger struct {
 }
 
 func (sl *stdLogger) Log(lvl int, msg string) {
+	if sl.lInfo.DebugLvl <= 0 {
```
Gilthoniel, line 133: A comment there to explain why you're printing although the level is below 0 could help future readers (and me by the occasion)

PR #637, opened 2020-04-28
Better tracing defaults

ineiti:
By default, don't print the single spans. Also removes a remaining
debug-print. Adds TRACING_DEBUG environmental variable.

OR-review: first one to agree, please merge.

### DIFF
--- a/tracing/logger.go
+++ b/tracing/logger.go
@@ -95,7 +95,6 @@ func (logger *TraceLogger) AddOnetDefaults(si *network.ServerIdentity) {
 
 	logger.defaultFields["nodeName"] = si.String()
 	logger.defaultFields["nodeDescription"] = si.Description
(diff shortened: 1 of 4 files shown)


tracing/logger.go
```diff
@@ -196,6 +195,14 @@ func (logger *TraceLogger) AddEnvironment() error {
 		}
 		logger.NoSingleSpans = tcss == "true"
 	}
+	tdb := strings.ToLower(os.Getenv("TRACING_DEBUG"))
```
gnarula, line 198: A small nit but perhaps add `TRACING_DEBUG` to the method comment?

ineiti: Documentation is more than a nitpick!

ineiti: Also added to README.md

PR #640, opened 2020-05-14
Add filtering for input and output connections

### LINKED ISSUE #620: Limit allowed connections
To further lock down the nodes, it should be possible to define which nodes are allowed to connect. This can be a simple extension of the `Server` structure with something like:

`AllowNodes([]network.ServerIdentityID)`

Whenever a new block is accepted by the byzcoin service, it would call this method with the latest roster.

The `AllowRoster` should limit both incoming and outgoing connections to be restricted to the set given in `AllowNodes`.

An environmental variable can be used to switch between the behaviour before `AllowNodes` is called:
- `locked` - nothing allowed
- `open` - everybody can connect
- `xxxx:xxxx:....` - a list of ServerIdentityIDs allowed to connect

cgrigis:
This PR implements a filtering of the input and output connections, allowing to mitigate possible DoS/amplification attacks.

It adds a data structure in `Router` that manages the valid peers, organized in subsets identified by a key. This allows to group the subsets by a logical entity (e.g. a service), which can evolve individually. For example, the ByzCoin service will update a set (specific to the ByzCoin service and a SkipChainID) every time a new block is added, with the peers corresponding to the added block's roster.

The data structure is used before the router opens an outgoing connection or accepts an incoming connection. At the router level, a peer is valid if it is present in any of the subsets (i.e. if it belongs to the union of the subsets).

See #620 and dedis/cothority#2204 for the motivation.

Closes #620

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -124,7 +124,7 @@ func (r *Router) Start(isPeerValid func(ServerIdentityID) bool) {
 			log.Errorf("rejecting incoming connection from %v: invalid peer %v",
 				c.Remote(), dst.ID)
 			if err := c.Close(); err != nil {
-				log.Error("closing connection:", err)
+				log.Warnf("closing connection: %v", err)
 			}
 			return
 		}
--- a/network/router.go
+++ b/network/router.go
@@ -58,6 +58,76 @@ type Router struct {
 	UnauthOk bool
 	// Quiets the startup of the server if set to true.
 	Quiet bool
+
+	// Set of valid peers, used to filter allowed in/out connections.
+	// It is organized a data structure allowing for subsets of peers to evolve
+	// indipendently, each subset being identified by a PeerSetID.
+	validPeers validPeers
+}
+
+// PeerSetID is the identifier for a subset of valid peers.
+// This should typically be linked in a unique way to a service, e.g.
+// hash(serviceID | skipChainID) for ByzCoin
+type PeerSetID [32]byte
+
+// peerSet is the type representing a subset of valid peers, implemented as a
+// map of empty structs.
+type peerSet map[ServerIdentityID]struct{}
+
+// validPeers is the type representing all the valid peers.
+type validPeers struct {
+	peers map[PeerSetID]peerSet
+	lock  sync.Mutex
+}
+
+// Sets the set of valid peers for a given identifier.
+func (vp *validPeers) set(peerID PeerSetID, peers []*ServerIdentity) {
+	newPeers := make(peerSet)
+
+	for _, peer := range peers {
+		newPeers[peer.ID] = struct{}{}
+	}
+
+	vp.lock.Lock()
+	defer vp.lock.Unlock()
+
+	if vp.peers == nil {
+		vp.peers = make(map[PeerSetID]peerSet)
+	}
+
+	vp.peers[peerID] = newPeers
+}
+
+// Checks whether the given peer is valid (among all the subsets).
+func (vp *validPeers) isValid(peer *ServerIdentity) bool {
+	vp.lock.Lock()
+	defer vp.lock.Unlock()
+
+	// Until valid peers are initialized, all peers are valid
+	if vp.peers == nil {
+		return true
+	}
+
+	// Search whether the given peer is valid in any of the peer subsets
+	for _, peers := range vp.peers {
+		_, ok := peers[peer.ID]
+		if ok {
+			return true
+		}
+	}
+
+	return false
+}
+
+// SetValidPeers sets the set of valid peers for a given PeerSetID
+func (r *Router) SetValidPeers(peerID PeerSetID, peers []*ServerIdentity) {
+	r.validPeers.set(peerID, peers)
+}
+
+// isPeerValid checks whether the provided ServerIdentity is among the valid
+// peers for this router.
+func (r *Router) isPeerValid(peer *ServerIdentity) bool {
+	return r.validPeers.isValid(peer)
 }
 
 // NewRouter returns a new Router attached to a ServerIdentity and the host we want to
@@ -97,7 +167,7 @@ func (r *Router) Unpause() {
 
 // Start the listening routine of the underlying Host. This is a
 // blocking call until r.Stop() is called.
-func (r *Router) Start(isPeerValid func(ServerIdentityID) bool) {
+func (r *Router) Start() {
 	if !r.Quiet {
 		log.Lvlf3("New router with address %s and public key %s", r.address, r.ServerIdentity.Public)
 	}
@@ -120,7 +190,7 @@ func (r *Router) Start(isPeerValid func(ServerIdentityID) bool) {
 		}
 
 		// Reject incoming connections from invalid peers
-		if !isPeerValid(dst.ID) {
+		if !r.isPeerValid(dst) {
 			log.Errorf("rejecting incoming connection from %v: invalid peer %v",
 				c.Remote(), dst.ID)
 			if err := c.Close(); err != nil {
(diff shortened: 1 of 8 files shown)


network/router.go
```diff
@@ -118,6 +118,17 @@ func (r *Router) Start() {
 			}
 			return
 		}
+
+		// Reject incoming connections from invalid peers
+		if !isPeerValid(dst.ID) {
+			log.Errorf("rejecting incoming connection from %v: invalid peer %v",
```
ineiti, line 124: That sounds more like a `log.Warnf` to me - let's take @tharvik 's vote to decide ;)

network/router.go
```diff
@@ -118,6 +118,17 @@ func (r *Router) Start() {
...
+				c.Remote(), dst.ID)
+			if err := c.Close(); err != nil {
+				log.Error("closing connection:", err)
+			}
+			return
```
ineiti, line 129: Why doesn't this return an error?

network/router_test.go
```diff
@@ -451,6 +459,48 @@ func testRouterSendMsgDuplex(t *testing.T, fac routerFactory) {
...
+	require.NoError(t, err)
+	r2, err := NewTestRouterTCP(7879)
+	require.NoError(t, err)
+
+	// r1 does not accept any connection
+	go r1.Start(peerIsNeverValid)
+
+	log.OutputToBuf()
+	_, err = r2.Send(r1.ServerIdentity, &SimpleMessage{42})
+	require.Error(t, err)
```
ineiti, line 473: If this test fails, then all other tests will have `OutputToBuf`, which is not ideal. Perhaps a `defer log.OutputToOs()`?

server.go
```diff
@@ -270,3 +275,53 @@ func (c *Server) callTestClose() {
 	c.serviceManager.servicesMutex.Unlock()
 	wg.Wait()
 }
+
+// SetValidPeers sets the set of peers with which this server can communicate.
+func (c *Server) SetValidPeers(roster *Roster) {
```
ineiti, line 280: As above, use `[]*network.ServerIdentity` for minimal interface.

server.go
```diff
@@ -270,3 +275,53 @@ func (c *Server) callTestClose() {
...
+	newPeers := make(map[network.ServerIdentityID]struct{})
+
+	for _, node := range roster.List {
+		newPeers[node.ID] = struct{}{}
```
ineiti, line 284: `ID` is deprecated

more `ID`s in later code - not commented

ineiti: TIL - https://stackoverflow.com/questions/37320287/maptstruct-and-maptbool-in-golang

So I should get rid of my `map[]bool`...

network/router.go
```diff
@@ -97,7 +97,7 @@ func (r *Router) Unpause() {
 
 // Start the listening routine of the underlying Host. This is a
 // blocking call until r.Stop() is called.
-func (r *Router) Start() {
+func (r *Router) Start(isPeerValid func(ServerIdentityID) bool) {
```
ineiti, line 100: This would break the public API of `Router`, and I don't see an imperative reason to do so.
Can you rather keep the `isPeerValid` as an internal field that can be overwritten by the tests?
Or use a `SetPeerValidation` method to change the default peer validation.
Or add a `StartWithPeerValidation`, then have `Start` call `StartWithPeerValidation(r.defaultPeerValidation)`?

network/router.go
```diff
@@ -118,6 +118,17 @@ func (r *Router) Start() {
 			}
 			return
 		}
+
+		// Reject incoming connections from invalid peers
+		if !isPeerValid(dst.ID) {
```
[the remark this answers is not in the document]
ineiti: `Roster.ID` is deprecated - `ServerIdentity.ID` is OK...

network/router.go
tharvik: indeed, more like a warning.

network/router.go
cgrigis: I found it cleaner this way, but indeed it breaks `Router`'s public API.
I will change it.

network/router.go
cgrigis: OK, for my future reference, what is the criteria? :)

network/router.go
cgrigis: Because `Start()` does not return one (called as a goroutine).

network/router_test.go
cgrigis: I have a few more issues with this test:
* The `require` on line #473 fails as is written here when I run it locally (i.e. `err == nil`), but on Travis it usually gets `err != nil` (but not always...).
* Furthermore, having to `Sleep()` and `Stop()` the routers to get the result feels very brittle to me (how long should I `Sleep` for?).

Would you have a suggestion as to how to better write these tests?

network/router.go
tharvik: IMO, it's an expected behavior, a interesting-to-see failure but not one that would normally be acted upon, compared to an "error" that should be investigated.

cgrigis: If everything is working as it should, it is not really an expected behavior (maybe indicative of an attack).
But I agree it is a fine line... I will bow to the majority ;-) .

network/router.go
ineiti: But this runs in `Listen`, which does return an error, no?

cgrigis: Indeed not `Start()`, but as I understand it, it actually runs in the function argument to `Listen()`, which does not return an error. What am I missing?

network/router.go
ineiti: Writing this down makes me think there are a lot of errors that should be warnings ;) I had to add `panic` to the list to think straight:

- `Panic` - if the program continues, it might make things worse: db-failure, out-of-memory error, consistency-checks failure that cannot be recovered from, ...
- `Error` - an action **initiated by the system** couldn't be executed, but it's not critical. Probably mostly network failures: couldn't contact a node, invalid reply from the network
- `Warning` - something noteworthy arrived at the system that might be interesting to check out: invalid (connection) request, failed transaction

Like always - once you actually start thinking about it, it's not that easy ;)

For me the difference between error and warning is the initiator: an error is something that the system tried to do and failed, while a warning is a failure while trying to execute an action from the outside.

Which would clean up a lot the failing transactions that are currently `Error`s, but according to this list, they would be `Warning`s. Which make them less: `DO SOMETHING ABOUT IT - NOW!!`

network/router.go
ineiti: I was really looking hard at the code and thought I'm probably wrong - but I didn't see it until now.  `r.host.Listen` returns an error, but not the function given in the method. Sorry for the noise...

network/router_test.go
ineiti: OK, played around with it a bit, and I think the problem is that the `Router.Send` method does a `r.connect`, which it thinks it succeeds, as the connection is only closed _after_ the handshake.
I tried to close the connection in `Router.receiveServerIdentity`, but that doesn't work, neither. 

So instead of testing the return error of `Router.Send`, which is not well defined, ignore the error from `r2.Send` and check if `r1.connections` has an entry for `r2`.

ineiti: Which means that you have to wait for the error in the `StdErr` of `r1`...

cgrigis: Thank you!

network/router.go
cgrigis: I agree with the `Panic`, which pertains to situations from which the software cannot recover.
For me. however, the difference between `Error` and `Warning` is not so much related to the initiator, but to the severity of what happened, a warning indicating an unexpected situation that might or might not lead to something more serious in the future, and an error being a situation already serious.

But as always, the most important thing is to have clear definitions to ensure objectivity in the choices. :)

Examples of common logging libraries that I am familiar with and how they define their levels:
* Python standard `logging` library: https://docs.python.org/3.8/howto/logging.html#logging-basic-tutorial
* `Log4j`, a popular logging library for java: https://en.wikipedia.org/wiki/Log4j#Log4j_log_levels

ineiti:
If you're happy with it, as this is onet, you can ask Gaylor or Jeff to give his final blessing for this PR.

PR #641, opened 2020-05-22
Set log-level to 3 for tracing

ineiti:
Reduce the amount of tracing that is sent to honeycomb,
as currently there are peaks that are not recorded.

After this PR is accepted, could you please also bump the onet version?

OR review - first one to agree, please merge ;)


nkcr:
By "bumping the Onet version" you mean adding a new git tag ?

ineiti:
yes, I think `v3.2.2` would be the next one. Then I'll create a PR for dedis/cothority to point to the new onet, and finally have my log-bug out of the way that doubles all logging messages when DEBUG_LVL=0...

PR #648, opened 2020-06-16
Minor fixes found while debugging

jeffallen:
These are minor changes I made while I was looking into failures on DETERLab yesterday.

### DIFF
--- a/simul/platform/cliutils.go
+++ b/simul/platform/cliutils.go
@@ -56,6 +56,7 @@ func SSHRun(username, host, command string) ([]byte, error) {
 		addr = username + "@" + addr
 	}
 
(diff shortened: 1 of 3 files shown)


simul/platform/cliutils.go
```diff
@@ -56,6 +56,7 @@ func SSHRun(username, host, command string) ([]byte, error) {
 		addr = username + "@" + addr
 	}
 
+	log.Lvl4("Going to ssh to", addr, command)
```
ineiti, line 59:
```suggestion
	log.Lvl2("Going to ssh to", addr, command)
```
I would put this even in Lvl2, seems quite interesting ;)

PR #656, opened 2020-08-20
Got it to failing, but...

ineiti:
The update to the new TLS handshake is faulty. The test in #655 only tests old->old and new->new, but not new->old and old->new (which fails).

This PR tests all combinations, but fails. Currently I don't know why :(

The relevant test is here: https://github.com/dedis/onet/pull/656/files#diff-f183c5e9370db10aaa3413143732a5abR46

### DIFF
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -278,7 +278,7 @@ func handleError(err error) error {
 
 	netErr, ok := err.(net.Error)
 	if !ok {
-		return xerrors.Errorf("non-network error: %v", err)
+		return ErrUnknown
 	}
 	if netErr.Timeout() {
 		return ErrTimeout
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -116,7 +116,7 @@ func NewTCPConn(addr Address, suite Suite) (conn *TCPConn, err error) {
 func (c *TCPConn) Receive() (env *Envelope, e error) {
 	buff, err := c.receiveRaw()
 	if err != nil {
-		return nil, xerrors.Errorf("receiving: %w", err)
+		return nil, xerrors.Errorf("receiving: %v", err)
 	}
 
 	id, body, err := Unmarshal(buff, c.suite)
@@ -278,7 +278,7 @@ func handleError(err error) error {
 
 	netErr, ok := err.(net.Error)
 	if !ok {
-		return ErrUnknown
+		return xerrors.Errorf("non-network error: %v", err)
 	}
 	if netErr.Timeout() {
 		return ErrTimeout
--- a/network/tcp.go
+++ b/network/tcp.go
@@ -116,7 +116,7 @@ func NewTCPConn(addr Address, suite Suite) (conn *TCPConn, err error) {
 func (c *TCPConn) Receive() (env *Envelope, e error) {
 	buff, err := c.receiveRaw()
 	if err != nil {
-		return nil, xerrors.Errorf("receiving: %v", err)
+		return nil, xerrors.Errorf("receiving: %w", err)
 	}
 
 	id, body, err := Unmarshal(buff, c.suite)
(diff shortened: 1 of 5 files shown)


network/tls.go
```diff
@@ -345,23 +351,30 @@ func makeVerifier(suite Suite, them *ServerIdentity) (verifier, []byte) {
...
-				for _, u := range cert.URIs {
-					if u.Scheme == "onet-pubkey" && u.Opaque == cn {
-						found = true
-						break
-					}
-				}
-				if !found {
-					return xerrors.Errorf("No onet-pubkey URIs match the expected public key %v", pubToCN(them.Public))
+			if noURIs {
+				err = cert.VerifyHostname(pubToCN(them.Public))
```
jeffallen, line 355: This call will not pass, no matter what, if you are testing with Go 1.15. The idea of this PR is to provide two implementations, one like v3.2.4 and one like v3.2.5 and mix and match them. That's cool, but then you can only get it to pass with Go 1.14.

I installed Go 1.14 and then ran the unit test like this:
```
brew install go@1.14
/usr/local/Cellar/go@1.14/1.14.7/bin/go test -run TestTLS -v
```

network/tls.go
```diff
@@ -184,6 +185,7 @@ func (cm *certMaker) get(nonce []byte) (*tls.Certificate, error) {
 	// test code to get ahold of and modify.
 	if testNoURIs {
 		tmpl.URIs = nil
+		tmpl.MaxPathLen = 1
```
jeffallen, line 188: This is necessary as a true repro of v3.2.4 behaviour, but it is not important. It will cause Go 1.15's server-side TLS to fail with an "internal error".

network/tls.go
```diff
@@ -306,7 +308,8 @@ type verifier func(rawCerts [][]byte, vrf [][]*x509.Certificate) (err error)
 // gives us a certificate back, crypto/tls calls the verifier with arguments we
 // can't control. But the verifier still has access to the nonce because it's in the
 // closure.
-func makeVerifier(suite Suite, them *ServerIdentity) (verifier, []byte) {
+func makeVerifier(suite Suite, them *ServerIdentity) (verifier,
```
jeffallen, line 311: This is bad formatting. If this is the fault of the pre-commit stuff, it needs to be fixed somehow.

network/tls.go
```diff
@@ -440,7 +450,8 @@ func pubToCN(pub kyber.Point) string {
 
 // tlsConfig returns a generic config that has things set as both the server
 // and client need them. The returned config is customized after tlsConfig returns.
-func tlsConfig(suite Suite, us *ServerIdentity) (*tls.Config, error) {
+func tlsConfig(suite Suite, us *ServerIdentity) (*tls.Config,
+	error) {
```
jeffallen, line 454: pre-commit, I hate you.

network/tcp.go
```diff
@@ -116,7 +116,7 @@ func NewTCPConn(addr Address, suite Suite) (conn *TCPConn, err error) {
 func (c *TCPConn) Receive() (env *Envelope, e error) {
 	buff, err := c.receiveRaw()
 	if err != nil {
-		return nil, xerrors.Errorf("receiving: %w", err)
```
jeffallen, line 119: I don't know why, but this change causes TestRouterErrorHandling to fail.

jeffallen:
So I'm a little unclear on where the changes to the error logging came from, don't know if they need to be there, in which case we need to fix the tests.

Also, not clear on what this tells us: it seems like this proves that this change is backwards compatible, but clearly the experience with the production network says it was not.

ineiti:
> Also, not clear on what this tells us: it seems like this proves that this change is backwards compatible, but clearly the experience with the production network says it was not.

What has been 'tested in prod' and failed is the following:
_old (1.14) -> new (1.15)_
And this is very hard to test - should be more of an integration test with a `test.sh` that takes _old_ code and compiles it with 1.14 and tries to connect to _new_ code compiled with 1.15...

Closing in favour of #657

PR #661, opened 2020-09-01
Replace satori/go.uuid with google/uuid

### LINKED ISSUE #660: go.uuid import is broken
After merging of #659 , it turns out that satori/go.uuid is broken and doesn't work as intended with a go-module. As at least one other fork of it, gofrs/uuid, removed go.mod support, I propose go.uuid is replaced with `google/uuid`.

In the meantime, a workaround can be

```
	github.com/satori/go.uuid v1.2.1-0.20181028125025-b2ce2384e17b // indirect
```

in onet/go.mod.

gnarula:
Replaces the uuid dependency to google/uuid which is actively maintained and uses go.mod

Closes #660

### DIFF
--- a/context.go
+++ b/context.go
@@ -6,7 +6,6 @@ import (
 	"encoding/binary"
 	"sync"
 
-	uuid "github.com/satori/go.uuid"
 	"go.dedis.ch/onet/v3/log"
 	"go.dedis.ch/onet/v3/network"
 	bbolt "go.etcd.io/bbolt"
@@ -331,7 +330,7 @@ func (c *Context) GetValidPeers(peerID network.PeerSetID) []network.
 func (c *Context) NewPeerSetID(data []byte) network.PeerSetID {
 	// Compute the PeerSetID as hash(serviceID | data)
 	h := sha256.New()
-	h.Write(uuid.UUID(c.ServiceID()).Bytes())
+	h.Write(c.serviceID[:])
 	h.Write(data)
 
 	return network.NewPeerSetID(h.Sum(nil))
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/golangplus/bytes v0.0.0-20160111154220-45c989fe5450 // indirect
 	github.com/golangplus/fmt v0.0.0-20150411045040-2a5d6d7d2995 // indirect
 	github.com/golangplus/testing v0.0.0-20180327235837-af21d9c3145e // indirect
-	github.com/google/uuid v1.1.1
+	github.com/google/uuid v1.1.2
 	github.com/gorilla/websocket v1.4.0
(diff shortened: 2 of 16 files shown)


ineiti (changes requested):
LGTM - one question remains:

- is it in fact OK to replace the uuid-library, as the uuid.UUID is used in some of the types? -> do we break the API with that change?

gnarula:
I don't think `uuid.UUID` is exposed directly to the users - the end users would always deal with types like `ServerIdentityID` and the methods defined on it.

Under the hood, both dependencies share the same representation:
https://github.com/satori/go.uuid/blob/b2ce2384e17bbe0c6d34077efa39dbab3e09123b/uuid.go#L37 and https://github.com/google/uuid/blob/0e4e31197428a347842d152773b4cace4645ca25/uuid.go#L19

ineiti:
Did you try to compile `dedis/cothority` using these changes?

ineiti (approved):

ineiti:
So I can compile `dedis/cothority` with these changes - so it should be OK...

gnarula:
@ineiti it builds fine but I'm getting some test failures on `dedis/cothority` which I'm looking into

>>> router.go:215 (network.(*Router).Start.func1) - receiving server identity from "" failed: Error while receiving ServerIdentity during negotiation receiving: [::1]:64212 sends too big packet: 1195725856>10485760

ineiti:
> Edit: Nvm, getting it irrespective of this change :/

Which test is that?

gnarula:
> > Edit: Nvm, getting it irrespective of this change :/
> 
> Which test is that?

`blscosi.TestClient_SignatureRequest`, `blscosi.TestService_SignatureRequest` `byzcoin.TestClient_NoPhantomSkipchain`,`contracts.TestDeferred_ExpireBlockIndex`, `calypso.TestClient_Authorize`

ineiti:
> `blscosi.TestClient_SignatureRequest`, `blscosi.TestService_SignatureRequest` `byzcoin.TestClient_NoPhantomSkipchain`,`contracts.TestDeferred_ExpireBlockIndex`, `calypso.TestClient_Authorize`

Here the error disappears if I run them individually - can you confirm that? Else we'll need to open an issue for this...

gnarula:
=

> > `blscosi.TestClient_SignatureRequest`, `blscosi.TestService_SignatureRequest` `byzcoin.TestClient_NoPhantomSkipchain`,`contracts.TestDeferred_ExpireBlockIndex`, `calypso.TestClient_Authorize`
> 
> Here the error disappears if I run them individually - can you confirm that? Else we'll need to open an issue for this...

Same here, errors disappear if I run the tests individually

ineiti:
So I think this should go in.

PR #664, opened 2020-11-05
Remove send restriction

ineiti:
The valid peer list should only be applied when receiving new messages.
Sending to any node should always be allowed, as this is something that
comes from the inside, and we suppose that the protocols are correct.

Will need to bump the tag to v3.2.6 for inclusion in cothority

### DIFF
--- a/network/router.go
+++ b/network/router.go
@@ -284,11 +284,6 @@ func (r *Router) Stop() error {
 // the messages are sent through the same connection and thus are correctly
 // ordered.
 func (r *Router) Send(e *ServerIdentity, msgs ...Message) (uint64, error) {
-	if !r.isPeerValid(e) {
-		return 0, xerrors.Errorf("%v rejecting send to %v: invalid peer",
-			r.ServerIdentity.ID, e.ID)
-	}
-
 	for _, msg := range msgs {
 		if msg == nil {
 			return 0, xerrors.New("cannot send nil-packets")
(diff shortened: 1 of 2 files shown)


cgrigis:
Just curious: I understand the reasoning, but even if we "suppose that the protocols are correct", I'd say that more checks are always better. :) Why is it needed? Is there a use-case that is blocked by this behavior?

ineiti:
Two reasons - a good and a not so good one ;)
1. think of it as your standard firewall that doesn't allow incoming connections, but allows all outgoing connections
2. as we applied this after the creation of all the protocols, some things are failing with this restriction, as the protocols don't correctly handle "ask before connecting to new nodes". 

Concrete case for 2.: I re-arranged how the blocks are handled for https://github.com/dedis/cothority/pull/2400 and it fails when the roster is updated, as the new roster is used before `SetValidPeers` is called.

cgrigis:
>   think of it as your standard firewall that doesn't allow incoming connections, but allows all outgoing connections

For the record, my home firewall does not allow all outgoing connections. ;-) 

>   as we applied this after the creation of all the protocols, some things are failing with this restriction, as the protocols don't correctly handle "ask before connecting to new nodes".
> 
> Concrete case for 2.: I re-arranged how the blocks are handled for [dedis/cothority#2400](https://github.com/dedis/cothority/pull/2400) and it fails when the roster is updated, as the new roster is used before `SetValidPeers` is called.

I see.

cgrigis (approved):

PR #676, opened 2021-04-14
Initialize ServerIdentities for local tests

ineiti:
While fixing a missing test in byzcoin, I saw that the ServerIdentities are not fully
initialized for local tests. As the new configuration contract in byzcoin also
checks for Address and Description in the ServerIdentity, they also must be
set for local tests.

Once this PR is merged, the onet version should be increased.

### DIFF
--- a/local.go
+++ b/local.go
@@ -487,7 +487,6 @@ func NewPrivIdentity(suite network.Suite, port int) (kyber.Scalar, *network.Serv
 	kp := key.NewKeyPair(suite)
 	id := network.NewServerIdentity(kp.Public, address)
 	ServiceFactory.generateKeyPairs(id)
-	id.Address = address
 	id.Description = fmt.Sprintf("LocalNode-%d", port)
 
 	return kp.Private, id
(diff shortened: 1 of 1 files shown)


local.go
```diff
@@ -487,6 +487,8 @@ func NewPrivIdentity(suite network.Suite, port int) (kyber.Scalar, *network.Serv
 	kp := key.NewKeyPair(suite)
 	id := network.NewServerIdentity(kp.Public, address)
 	ServiceFactory.generateKeyPairs(id)
+	id.Address = address
```
nkcr, line 490: Seems useless as this is already done in `NewServerIdentity`.

ineiti: The PR is getting very small ;)

nkcr (approved):

local.go
ineiti: But it's needed for https://github.com/dedis/cothority/pull/2445/files#diff-01980534bc5287f23e37d7f5af82156839e052f43a435b3da70f2a8a731aa8b6R684

PR #681, opened 2021-09-27
Test with go 1.17.x

ineiti:
Updates the tests so that they pass with go 1.17.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - "1.15.x"
+  - "1.17.x"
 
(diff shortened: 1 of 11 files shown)


ineiti:
Hmm - go 1.17 introduced a new way to add tags to the .go files. And `gofmt` insists on imposing these tags. But then it won't be compatible with go 1.16 anymore. But we want to have the latest two go-versions supported.

ineiti:
OK, `gofmt` only _adds_ a `//go:` line, so go 1.16 should be OK.

tharvik (approved):

